PR #1, opened 2022-07-07
add Tool category

doodleEsc:
Plugins such as [gopher.nvim](https://github.com/olexsmir/gopher.nvim) need bunch of binary tools which are not classicifed as `Linter` or `Formater`

We can use `Tool` category to manage those plugins


williamboman (commented):
Hello! Cool! I'm not too sure about adding a category called "Tool" as I feel like that's a bit too generic (wouldn't all packages belong under such a category?). I'd be fine with not specifying a category for this package just yet (then it'd only show up under `All` in the UI) - it can be an empty list for now (`lemmy-help` is another example of this)! 

The `scripts/autogen_metadata.sh` script is a bit unstable, it seems like it failed when you ran it causing it to delete a bunch of files. Try rerunning it again, if that doesn't work I can patch it up for you

PR #9, opened 2022-07-07
feat: add dart-debug-adapter

williamboman:


akinsho:
@williamboman I actually the nvim-dap entry quite a while ago since at the time that was the only way to get that installed, I was using it in my [dart (flutter) plugin](https://github.com/akinsho/flutter-tools.nvim), but thankfully it got built in natively so isn't needed any more. It's quite brittle since the author of that plugin didn't intend for it to be used this way, i.e. for non-vscode editors, so it was only out of desperation. I should probably have updated that at some point.

PR #15, opened 2022-07-08
add go tools

doodleEsc:
add a buncho of go tools under the category 'ALL'

1. [gotests](https://github.com/cweill/gotests): generate table driven tests for go function
2. [gomodifytags](https://github.com/fatih/gomodifytags): add tags for go struct
3. [impl](https://github.com/josharian/impl): generates method stubs for implementing an interface


williamboman (approved):
LGTM! You'll need to run the `script/autogen_metadata.sh` script for these packages to be indexed. I'd recommend running it like so, it shouldn't cause any issues:

```sh
$ SKIP_SCHEMAS=1 scripts/autogen_metadata.sh
```

I could've done this for you and pushed to the branch, but since you're using the default `alpha` branch I'll avoid doing so because by doing so, GitHub will start considering me responsible for your fork as well (I'll get notifications and stuff for it)

williamboman:
Superseded by #28, thanks @doodleEsc !

PR #31, opened 2022-07-08
feat: add eslint_d linter

Relms12345:
This PR adds support for the eslint_d linter

its the recommended linter by null-ls for js/ts files because it creates a eslint server in the background to make linting much faster compared to just using the default eslint command for every save. (speeds up linting by 400+%)

### DIFF
--- a/lua/mason-registry/eslint_d/init.lua
+++ b/lua/mason-registry/eslint_d/init.lua
@@ -3,7 +3,7 @@ local npm = require "mason-core.managers.npm"
 
 return Pkg.new {
     name = "eslint_d",
-    desc = [[Eslint_d is a very fast eslint server for JavaScript and TypeScript]],
+    desc = [[Makes eslint the fastest linter on the planet]],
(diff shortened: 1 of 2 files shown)


lua/mason-registry/eslint_d/init.lua
```diff
@@ -0,0 +1,11 @@
+local Pkg = require "mason-core.package"
+local npm = require "mason-core.managers.npm"
+
+return Pkg.new {
+    name = "eslint_d",
+    desc = [[Eslint_d is a very fast eslint server for JavaScript and TypeScript]],
```
williamboman, line 6:
```suggestion
    desc = [[Makes eslint the fastest linter on the planet]],
```

(I tend to source these descriptions directly from upstream)

williamboman (approved):
Awesome!

lua/mason-registry/eslint_d/init.lua
Relms12345: ah got it i'll remember that for next time :D

PR #45, opened 2022-07-10
fix(luarocks): bin linking

williamboman:


williamboman (commented):
@williambotman /generate

PR #46, opened 2022-07-10
feat: add luacheck to the registry

WhoIsSethDaniel:


williamboman (approved):
@williambotman /generate

PR #49, opened 2022-07-11
chore(lemminx): update latest default version

williamboman:


williamboman:
@williambotman /generate

williamboman:
@williambotman /generate pls

PR #59, opened 2022-07-12
chore!: remove ccls

williamboman:
This is a package that doesn't see frequent releases and it's been a
mess to get to build across platforms.

Installing via the system default package manager provides a much better
experience atm.


williamboman:
@williambotman /fixup

ritchielrez:
Though the decision is understandable, having `ccls` *lsp-installer* could have really helped us, the**Windows** users.

PR #66, opened 2022-07-13
fix(api): fix the :MasonUninstall command

williamboman:


williamboman:
@williambotman /fixup

PR #81, opened 2022-07-16
Add package

shaeinst:
Add djlint


lua/mason-registry/djlint/init.lua
```diff
@@ -0,0 +1,11 @@
+local Pkg = require "mason-core.package"
+local pip3 = require "mason-core.managers.pip3"
+
+return Pkg.new {
+    name = "djlint",
+    desc = [[HTML Template Linter and Formatter. Django - Jinja - Nunjucks - Handlebars - GoLang]],
+    homepage = "https://github.com/Riverside-Healthcare/djLint",
+    languages = { Pkg.Lang.HTMLDjango },
```
williamboman, line 8: I think `Django` would make more sense as a language. This is only used to aid with discovery, and my rationale is that you're probably more likely to look for tools for "Django" than for "HTMLDjango":

Also, looking at their README, I think more languages should be added?

```suggestion
    languages = { Pkg.Lang.Django, Pkg.Lang.Go, Pkg.Lang.Nunjucks, Pkg.Lang.Handlebargs, Pkg.Lang.Mustache, Pkg.Lang.Angular, Pkg.Lang.Jinja },
```

williamboman (commented):
Cool!

williamboman:
@williambotman /fixup

williamboman:
@williambotman /fixup

PR #90, opened 2022-07-17
feat(ui): bring focus to the installation section when :MasonInstall

williamboman:
This ensures users see the progress output immediately, instead of
having to scroll down to see it.


williamboman:
@williambotman /fixup

PR #100, opened 2022-07-19
fix(mason-lspconfig): patch some server's cmd on Windows

williamboman:
Bit dirty but I think this gives the best experience OOTB.


williamboman:
@williambotman /fixup

Issue #103, opened 2022-07-20
Configuration to auto install missing tools


alisonjenkins:
### Is your feature request related to a problem? Please describe.

Not a problem

### Describe the solution you'd like

If possible I would like a configuration option to allow passing a table of tools that I would like Mason to automatically install so that when I copy my config to a new machine that it will automatically install all the required tools for my configuration rather than me having to manually install each via the UI.

WhoIsSethDaniel:
Hi @alanjjenkins , you can try https://github.com/WhoIsSethDaniel/mason-tool-installer.nvim which does exactly this. I wrote it for the very use-case you mention.

alanjjenkins:
Fantastic thanks @WhoIsSethDaniel. 

Is it worth us putting this in the extensions area of the README for Mason to save someone else putting a duplicate issue in for this?

alanjjenkins:
> Fantastic thanks @WhoIsSethDaniel.
> 
> Is it worth us putting this in the extensions area of the README for Mason to save someone else putting a duplicate issue in for this?

Created PR: https://github.com/williamboman/mason.nvim/pull/104 for this.

williamboman:
I think I'll open up a Wiki page for these things! I will likely also extract the `mason-lspconfig` extension into its own plugin, to make boundaries clearer but also promote more people to write similar extensions like mason-dap and mason-null-ls (ideally such extensions wouldn't be required, but there are gaps to fill).

So to answer the original topic; I have been wanting to incorporate something like this into mason core, but before doing so I'd like to make sure it's solid and well thought out. Some concerns I have are:

- Need to think about how to handle new versions
  - Automatic checks for new versions need to be scheduled somehow, as to not do them every time you launch neovim
  - Immediately upgrading to a new version as soon as it's available will likely lead to some pretty disrupting experiences, should there be breaking changes in the new version. At the very least, should there be some mechanism that tries to parse versions and try to assess compatibility by conventional methods (semver etc.)?
- Need to think about how to handle short-lived neovim sessions - are there some heuristics we can use to decide to commit to a package upgrade (e.g, session has reached X seconds)
- This one's a bit easier, but need to figure out a nice outlet for displaying progress other than the `:Mason` window. Echoing stdout & stderr (like treesitter's `:TSUpdate`) would be one way. I've also wanted to explore windows like the one provided by https://github.com/j-hui/fidget.nvim (keeping an :eye: on https://github.com/j-hui/fidget.nvim/issues/70)

[...]

PR #104, opened 2022-07-20
Mention the existence of the meson-tool-installer plugin

alanjjenkins:
Adds a mention of the meson-tool-installer plugin for Meson for
automatic tool installation to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,3 +198,10 @@ require("mason-lspconfig").setup()
 ```
 
 For more documentation, see [`:h mason-lspconfig`](./doc/mason-lspconfig.txt).
+
+## `WhoIsSethDaniel/mason-tool-installer.nvim`
+
+`mason-tool-installer` is an extension to Meson that allows automatically installing and updating a list of configured plugins.
+
+This is useful if you wish to use your Neovim configuration across several machines and have the tools required for LSP
+and linting installed automatically.


williamboman (commented):
Thanks! Per my comment in #103, I've added [this to the Wiki instead](https://github.com/williamboman/mason.nvim/wiki/Extensions)!

etherTide:
I don't think that the wiki is available anymore, at least it isn't through that link. I've spent a long time looking through the :h mason page and the README trying to find a way to accomplish this. I was put on the right track going for an external plugin by a stroke of luck; my old config was kickstart.nvim which includes mason-tool-installer.

However, I really thought that this functionality would be in the mason API somewhere, specifically because of lines 122-125 in [doc/mason.txt](https://github.com/mason-org/mason.nvim/tree/main/doc/mason.txt):
```help
HOW TO INSTALL PACKAGES                        *mason-how-to-install-packages*

You may install packages either via the command interface or via Mason's Lua
APIs. See |:MasonInstall| for more details.
```
where I thought that "Mason's Lua APIs" meant that there would be some function or require().setup() option that I could add to my lua/plugins/mason.lua file that would ensure installation.

Reading this back, I'm worried that I'm sounding entitled and in no way mean to suggest that you must allow this functionality alla #103 or #483. I just wanted to say, after reading several issue threads where the solution in lieu of implementing the (erroneous) feature expectation was to clarify that users should look to other plugins, I think that it still isn't clear to users that they should look to other plugins.

PR #108, opened 2022-07-21
fix wrong configuration parameter

ofseed:
I found out that the type of parameter passed to the Volar configuration function is a table

<https://github.com/williamboman/mason.nvim/blob/f2a8cc435c6968dab63290491211cb74fdfe0c7d/lua/mason-lspconfig/init.lua#L75>

instead of a string. 

<https://github.com/williamboman/mason.nvim/blob/f2a8cc435c6968dab63290491211cb74fdfe0c7d/lua/mason-lspconfig/server_configurations/volar/init.lua#L4-L5>

The parameters required by this configuration function, `install_dir` of type string are placed in this table. This will cause some serious errors, such as Volar cannot find the `tsserverlibrary.js` it needs, and failing to start.

I fixed Volar in this PR now, but I think other LSP servers may also have this bug.

### DIFF
--- a/lua/mason-lspconfig/server_configurations/angularls/init.lua
+++ b/lua/mason-lspconfig/server_configurations/angularls/init.lua
@@ -2,8 +2,10 @@ local platform = require "mason-core.platform"
 local _ = require "mason-core.functional"
 local path = require "mason-core.path"
 
----@param install_dir string
-return function(install_dir)
+---@param config table
+return function(config)
+    local install_dir = config["install_dir"]
+
     local append_node_modules = _.map(function(dir)
         return path.concat { dir, "node_modules" }
     end)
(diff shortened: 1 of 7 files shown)


williamboman:
Ah, I changed it into a table recently without updating elsewhere 🤦‍♂️. Thanks, will look into other problematic locations!

ofseed:
Except for omnisharp, I have changed the parameter types of the LSP servers which may be incorrect.

williamboman:
Hey thanks @ofseed! This part of the code was just moved to a separate plugin, and this issue was fixed in [this commit](https://github.com/williamboman/mason-lspconfig.nvim/commit/152348ae54f1527b056bd63d1d9e6770e2a285b9)! Thanks for raising this!

PR #113, opened 2022-07-22
feat: minimize modules required by .setup(), lazily require functional modules

williamboman:
Reduces the modules required by mason when calling .setup() by 7, to a
total of 8.


williamboman:
@williambotman /fixup

PR #128, opened 2022-07-25
feat: add hadolint

thanhvule0310:


lua/mason-registry/hadolint/init.lua
```diff
@@ -0,0 +1,33 @@
...
+    categories = { Pkg.Cat.Linter },
+    ---@async
+    ---@param ctx InstallContext
+    install = function(ctx)
+        github
+            .download_release_file({
+                repo = "hadolint/hadolint",
+                out_file = platform.is.win and "hadolint.exe" or "hadolint",
+                asset_file = coalesce(
+                    when(platform.is.mac_x64, "hadolint-Darwin-x86_64"),
```
williamboman, line 23: Mac x64 binaries can often be executed on the ARM chips as well! It seems like this one can, so let's do:

```suggestion
                    when(platform.is.mac, "hadolint-Darwin-x86_64"),
```

williamboman (approved):
🚀

PR #132, opened 2022-07-25
feat(prettierd): Add prettierd to the registry

NICHTJ3:
Add prettierd as a faster alternative to prettier


NICHTJ3:
Just go bet to it 😂

PR #138, opened 2022-07-25
feat: add cmake-format

WilsonOh:


williamboman:
@williambotman /fixup

williamboman:
Renaming the package and adding more executables in #139

PR #145, opened 2022-07-25
feat(ui): pass kind=mason-lang to vim.ui.select

MunifTanjim:
To allow styling it differently.

Example from neovim core: https://github.com/neovim/neovim/blob/7961f79904f6583ed3279fb1c73f2107ce9603ef/runtime/lua/vim/lsp/buf.lua#L807-L809

### DIFF
--- a/lua/mason/ui/instance.lua
+++ b/lua/mason/ui/instance.lua
@@ -546,6 +546,7 @@ end
 local function filter()
     vim.ui.select(_.sort_by(_.identity, _.keys(Package.Lang)), {
         prompt = "Select language:",
+        kind = "mason-lang"
     }, function(choice)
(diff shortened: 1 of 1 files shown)


lua/mason/ui/instance.lua
```diff
@@ -546,6 +546,7 @@ end
 local function filter()
     vim.ui.select(_.sort_by(_.identity, _.keys(Package.Lang)), {
         prompt = "Select language:",
+        kind = "mason-lang"
```
williamboman, line 549:
```suggestion
        kind = "mason.ui.language-filter"
```

williamboman (approved):
Ah TIL, cool!

williamboman:
@williambotman /fixup

PR #150, opened 2022-07-25
Add sqlfluff

dbernheisel:
Adds sqlfluff formatter and linter

For those using null-ls, you'll need to add extra flags when adding it, eg:

```lua
null_ls.builtins.formatting.sqlfluff.with({
  extra_args = {"--dialect", "postgres"}
})
```


williamboman:
@williambotman /fixup

PR #151, opened 2022-07-25
Add write good

dbernheisel:


williamboman (approved):

williamboman:
@williambotman /fixup

williamboman:
How maintained/relevant is this package? It seems like it's supported by null-ls so I'll go ahead and merge, but it seems somewhat abandoned.

dbernheisel:
great question, I didn't even notice that it's been months/years since its last update; but maybe it's one of those packages that doesn't need to be updated often (aka, it's "done"). It still seems relevant and works, so :shrug: and I use it.

PR #154, opened 2022-07-26
Provide direct copy paste snippet to run mason

beauwilliams:
I thought this would be good to have in readme to make it that little bit easier for new users to get up and running

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,11 @@ your personal usage, some of these will also need to be installed. Refer to `:ch
 ## [Packer](https://github.com/wbthomason/packer.nvim)
 
 ```lua
-use { "williamboman/mason.nvim" }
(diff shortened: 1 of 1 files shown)


RoryNesbitt:
This seems redundant since the very next thing shows the require. Unless that was removed and something similar was shown for vim-plug, therefore combining installation and setup sections

williamboman (commented):
Hey! I thought about doing this but decided not to with the following reasoning:

- it really isn't much code to get up and running (1 line for defaults)
- there are more alternatives than the `config` hook
- if someone doesn't know how to use packer, they'd only benefit from having to go through the process of setting up a plugin themselves a couple of times 😀

I did a quick random sample of other plugins and almost none do this. Thanks for the PR though!

PR #155, opened 2022-07-26
Add gitlint

thetic:


lua/mason-registry/gitlint/init.lua
```diff
@@ -0,0 +1,11 @@
+local Pkg = require "mason-core.package"
+local pip3 = require "mason-core.managers.pip3"
+
+return Pkg.new {
+    name = "gitlint",
+    desc = [[Gitlint is a git commit message linter written in python: it checks your commit messages for style.]],
+    homepage = "https://github.com/jorisroovers/gitlint",
```
williamboman, line 7: I tend to prefer to use a website if it's available, which it seems it is!

```suggestion
    homepage = "https://jorisroovers.com/gitlint/",
```

williamboman (approved):
Nice!

PR #160, opened 2022-07-26
Add shellharden

oncomouse:
Add [shellharden](https://github.com/anordal/shellharden) linter / formatter


williamboman:
@williambotman /fixup

williamboman:
@williambotman /fixup i shouldnt have to run this twice

PR #162, opened 2022-07-26
feat: add_package() for mason-registry

oncomouse:
In testing some new package additions to the registry, I found it frustrating that unless a package is added to `mason-registry.index`, things like `Package:is_installed()` don't seem to work. This adds a function to `mason-registry` that extends `mason-registry.index` and allows for locally authored packages to work as intended with `mason`.

Ideally, I'd like to be able to add a PackageSpec to index, but given the way index works, this seemed like the easiest solution for the moment.

### DIFF
--- a/lua/mason-registry/init.lua
+++ b/lua/mason-registry/init.lua
@@ -59,6 +59,15 @@ function M.is_installed(package_name)
     return scan_install_root()[package_name] == true
 end
 
+-- Add a require path to the index
+-- @param package_name string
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Hey, thanks for the PR! I've decided not to expose this kind of functionality just yet as I want things to settle first. Also need to think about how to deal with conflicting package names and how to integrate these custom packages into other parts.

PR #164, opened 2022-07-26
Fix shellharden install so that a link is created

WhoIsSethDaniel:


williamboman:
Nice catch! Reeeally need tests on a package-level. https://github.com/williamboman/nvim-lspconfig-test was pretty useful but very heavy to run and manage.

PR #196, opened 2022-07-29
fix(ui): ensure window is active before getting cursor

williamboman:


williamboman:
@williambotman /fixup

PR #201, opened 2022-07-30
feat(taplo): use prebuilt binary if available

williamboman:
I don't like this kind of mixing of installation sources, but I figured
this could be a good first experiment.

### DIFF
--- a/lua/mason-registry/taplo/init.lua
+++ b/lua/mason-registry/taplo/init.lua
@@ -1,5 +1,10 @@
 local Pkg = require "mason-core.package"
 local cargo = require "mason-core.managers.cargo"
+local github = require "mason-core.managers.github"
+local _ = require "mason-core.functional"
+local platform = require "mason-core.platform"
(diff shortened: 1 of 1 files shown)


williamboman:
Ping @zhang-stephen, this should be smoother now for Linux and Mac enjoyers

zhang-stephen:
well, I don't think pre-built is necessary, I am just confused about the difference between nvim-lsp-installer and mason yesterday. 

and I noticed the author of taplo seems do not like to "release", so building taplo from latest master is a more reasonable choice.

So this PR is not necessary to be merged in my opinion... What do you think about it?

zhang-stephen:
OK it's merged so quickly... 😂

williamboman:
> and I noticed the author of taplo seems do not like to "release", so building taplo from latest master is a more reasonable choice.

`cargo install` doesn't build from HEAD anyway, it goes through crates.io

PR #210, opened 2022-08-01
Add default option to highlights

thetic:
This avoids overriding user-provided highlight group definitions.
Discussed in #133.

https://neovim.io/doc/user/syntax.html#:hi-default


williamboman (approved):
Cool, seems like a good approach!

PR #211, opened 2022-08-01
feat: add spectral-language-server

ghassan0:
Add [spectral](https://github.com/luizcorreia/spectral-language-server) which has configs available at [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#spectral).


williamboman:
@williambotman /fixup

williamboman (approved):
Cool! That npm package is just a repackaged wrapper around a component of their official VSCode extension, similar to the HTML, CSS, ESLint, and JSON servers. I'd prefer if the original authors distributed this as a standalone component but for now this is what's readily available.

PR #252, opened 2022-08-05
feat(api): add MasonInstallComplete autocmd event

kylo252:
## Description

add an autocmd-event that will trigger when the installer is done

the callback takes `{name: string, success: boolean, result: table }`

example usage:

```lua
vim.api.nvim_create_autocmd("User", {
  pattern = "MasonInstallComplete",
  callback = function(args)
    vim.notify("MasonInstallComplete was called!, got: " .. vim.inspect(args.data))
  end,
})
```

semi-related: https://github.com/williamboman/mason.nvim/issues/169

### DIFF
--- a/lua/mason-core/package/init.lua
+++ b/lua/mason-core/package/init.lua
@@ -108,6 +108,12 @@ function Package:install(opts)
                 ---@param success boolean
                 ---@param result Result
                 function(success, result)
+                    vim.schedule(function()
+                        vim.api.nvim_exec_autocmds("User", {
+                            pattern = "MasonInstallComplete",
+                            data = { name = self.name, success = success, results = result },
+                        })
+                    end)
                     if not success then
                         log.error("Unexpected error", result)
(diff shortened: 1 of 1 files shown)


williamboman:
Hello! Hm what is the use case for this?

williamboman:
Aah I see. Yeah those fast events will get ya. I realize my small example is also subject to this 🥲. In the Mason source there's quite a few of:

```lua
if vim.in_fast_event() then
    a.scheduler()
end
```

and `vim.schedule_wrap`'s. For the time being I've taken the stance that it's the responsibility of the caller to schedule methods that cannot be called within a fast event. I'll make sure to highlight this somewhere [in the reference](https://github.com/williamboman/mason.nvim/blob/main/doc/reference.md).

So, for posterity, here's a slightly modified version of my previous example:

```lua
registry:on("package:install:success", vim.schedule_wrap(function (pkg)
    print(pkg.name .. " was installed")
end))
```

PR #254, opened 2022-08-05
fix(bash-debug-adapter): fix associated languages

williamboman:


Issafalcon:
@williamboman - Just seen this PR (ignore my question on the other one!). This makes sense now. It's what I originally thought would be the correct way to do it, but I forgot to modify the languages array when submitting the original PR. Thanks for fixing!

P.S. Great work on this project! I'm learning a lot about lua programming and plugin dev in general from reading through your plugins!

PR #256, opened 2022-08-06
debugpy load module `debugpy.adapter` instead of `debugpy`

eeeXun:
In the example https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation#python, we load module `debugpy.adapter` not `debugpy`

### DIFF
--- a/lua/mason-core/installer/context.lua
+++ b/lua/mason-core/installer/context.lua
@@ -217,7 +217,7 @@ end
 function InstallContext:write_pyvenv_exec_wrapper(new_executable_rel_path, module)
     return self:write_shell_exec_wrapper(
         new_executable_rel_path,
-        ("%q -m %s.adapter"):format(
+        ("%q -m %s"):format(
             path.concat {
                 require("mason-core.managers.pip3").venv_path(self.package:get_install_path()),
                 "python",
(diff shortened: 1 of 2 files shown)


williamboman:
Ah. I wonder if it'd make sense to just write another executable `debugpy-adapter` since both of these modules expose a different CLI interface? Might be nice to keep the main CLI linked as well?

eeeXun:
What about just use `exec "...mason/packages/debugpy/venv/bin/python"`, and let user write their arguments by themselves? So the dap configuration would be like

```lua
dap.adapters.python = {
  type = 'executable';
  command = 'debugpy';
  args = { '-m', 'debugpy' };  -- or { '-m', 'debugpy.adapter' }
}
```

williamboman:
Hm I personally find that very awkward, aliasing `debugpy` to `python`. I'm not sure why they don't provide wrapper executables themselves. I'd like to keep the `debugpy` executable as-is though, the question is how to expose the `debugpy.adapter` CLI.

eeeXun:
Oh, I see. I think add another `debugpy-adapter` make sense for me.

williamboman (approved):
Cool!

PR #268, opened 2022-08-08
docs: change to a more appropriate word

rafamadriz:
In this case, using `packages` instead of `languages` makes more sense.
In the example given, we're not installing just "languages" but also
other packages (a formatter)


williamboman (approved):
Ah, I totally did that on purpose to see if anyone reads the docs! You sir are a scholar and a gentleman

PR #274, opened 2022-08-09
don't download x64 taplo binary for arm64 mac

LC0117:
`taplo` only provides a x64 pre-compiled binary for mac, so it would be better to use cargo to install taplo on arm64 mac.

### DIFF
--- a/lua/mason-registry/taplo/init.lua
+++ b/lua/mason-registry/taplo/init.lua
@@ -16,7 +16,7 @@ return Pkg.new {
     ---@param ctx InstallContext
     install = function(ctx)
         local asset_file = coalesce(
-            when(platform.is.mac, "taplo-full-x86_64-apple-darwin-gnu.tar.gz"),
+            when(platform.is.mac_x64, "taplo-full-x86_64-apple-darwin-gnu.tar.gz"),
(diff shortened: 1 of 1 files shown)


williamboman:
Hello! Is this causing any issues, or is it mostly due to `amd64 != arm64`? The `amd64` binary should run fine on ARM thanks to Rosetta translation.

LC0117:
No big problem with the x64 binary aside from a little bit longer fresh-start time.
My original thought is that a mason package should be compatible with platforms, but considering that using cargo needs a rust toolchain, it would be more convenient to download from a release. I'll close this if you have the same opinion on the convenience that direct download can bring.

williamboman:
Yeah I figured any performance penalties are fully negligible regarding using compatible amd64 binaries on ARM Macs. Of course arm binaries are used when available, but that's not always the case. I also believe the translation is cached, so it only applies to the very first time you run the binary.

Avoiding the Cargo dependency for a package that has no relation to the Rust ecosystem has much more benefits than avoiding a Rosetta-translated binary (based on my somewhat limited understanding of Rosetta), so let's close this imo. Thanks for the PR though!

PR #285, opened 2022-08-11
feat(ui): check for new package versions when opening :Mason

williamboman:
This is enabled by default and can be disabled in settings.


williamboman:
@williambotman /fixup

PR #314, opened 2022-08-19
test(installer): add spec files for context and linker

williamboman:
- feat(installer): don't write exec wrappers for targets that don't exist
- test(installer): add spec files for context and linker


williamboman:
Fixed in #315!

PR #319, opened 2022-08-21
feat(csharp_ls): add csharp_ls support

Decodetalkers:
Log: csharp_ls support


williamboman:
Thanks! However, this package already exist under the naem [`csharp-language-server`](https://github.com/williamboman/mason.nvim/blob/main/PACKAGES.md#csharp-language-server)

PR #336, opened 2022-08-24
fixes help menu mapping mismatch

3699394:
Screenshot 2022-08-24 at 20 03 37

Help menu mapping did not change in https://github.com/williamboman/mason.nvim/pull/326


williamboman (approved):
👍

PR #366, opened 2022-09-01
feat: add luau-lsp

williamboman:


chubbymomo:
Hello, sorry for the off-topic question, but do you happen to have a config that makes this lsp work with lspconfig? I am rather new to customizing Nvim, and don't really know where to start with that.

Thank you very much!

adelarsq:
@chubbymomo Hi. There is a plugin for this. See https://github.com/williamboman/mason-lspconfig.nvim

PR #370, opened 2022-09-02
Bugfix for the fantomas binany name

adelarsq:
Bugfix for the fantomas binary name.

Tested with `:MasonInstall fantomas@5.0.0-beta-009`


williamboman (approved):
👍

PR #399, opened 2022-09-09
feat: add phpcs, phpcbf, phpmd, phpstan and twigcs

laytan:
Disclaimer: I did not test this on Windows, but the install functions are all a copy of the one for php-cs-fixer, so I am sure they will work on Windows.


williamboman (approved):
> so I am sure they will work on Windows.

Famous last words!

I tested on Windows, works fine 👍

williamboman (commented):
Actually, I wonder if `phpcbf` and `phpcs` should really be one package? I can't see a reason to separate these? I also wonder if it'd be best to simply use composer to install it, like so:

```lua
install = composer.packages { "squizlabs/php_codesniffer", bin = { "phpcs", "phpcbf" } }
```

My understanding is that composer is very ubiquitous among PHP developers, so imposing such a requirement is no big deal? The package name in that case could be `php_codesniffer`

laytan:
Yes composer is probably used by 99% of the php developers.

One drawback of using composer here is that if phpcs (for example) depends on another package, and the user or another package also depends on that package but another version of it, the installation will fail.

So personally, I like the standalone .phar here.

Also, the phpcs-fixer package is installed like I do in this ticket, so it might be good to keep consistent.

Let me know what you think, I am fine with changing them to use composer.

williamboman:
> One drawback of using composer here is that if phpcs (for example) depends on another package, and the user or another package also depends on that package but another version of it, the installation will fail.

Not sure I'm following. Composer installations in Mason are isolated and don't (or at least shouldn't) share dependencies with anything else.

Let's go with separate packages, I think I might've misunderstood the relation between these two packages

laytan:
Oh, my bad then, I thought the composer.packages way would use the composer `composer --global require` but you are doing it differently.

Anyway, thanks for the Mason and the merge!

PR #404, opened 2022-09-11
fix(ui): lower zindex to 45

williamboman:


guxuerui:
Sorry, So why did you change the `zindex` value from 50 to 45?  

Because I couldn't see the floating window when I was using `:Mason`, and when I changed the `zindex` value to be greater than 45 in my local source code, I could see it.

williamboman:
It was done for better interop with other 3rd party plugins that create floating windows. Hmm it sounds really weird if this would case the window to not be visible, unless you have a bunch of other floating windows open at the same time

guxuerui:
> It was done for better interop with other 3rd party plugins that create floating windows. Hmm it sounds really weird if this would case the window to not be visible, unless you have a bunch of other floating windows open at the same time

guxuerui:
> It was done for better interop with other 3rd party plugins that create floating windows. Hmm it sounds really weird if this would case the window to not be visible, unless you have a bunch of other floating windows open at the same time



I confirmed that only mason's window is open, anyway, I have now returned the zindex value to 99 and the window now displays

guxuerui:
And, thank you for the plugin~

PR #427, opened 2022-09-15
Pyright python package

### LINKED ISSUE #424: [New package]: pyright-python
### Package name

pyright-python

### Package homepage

https://github.com/RobertCraigie/pyright-python

### Languages

python

pip

It is a wrapper **in Python** around the installation of `pyright`.

Since the `pyright` LSP server is a tool for those who develop in Python and makes more sense, as a Python developer, to have all the tooling needed already installed (i.e. Python + pip), instead of the tooling for a language we are not developing in (i.e. Node + npm); it becomes obvious why this package is the first choice to provide `pyright`s LSP completion when developing in python for those who never need TypeScript.

baco:
Closes #424 feature request for adding new LSP package

### DIFF
--- a/PACKAGES.md
+++ b/PACKAGES.md
@@ -151,6 +151,7 @@
 - [pylint](#pylint)
 - [pyproject-flake8](#pyproject-flake8)
 - [pyright](#pyright)
+- [pyright-python](#pyright-python)
 - [python-lsp-server](#python-lsp-server)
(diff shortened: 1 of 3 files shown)


williamboman:
Closing per comment in https://github.com/williamboman/mason.nvim/issues/424#issuecomment-1248720683

PR #434, opened 2022-09-17
feat(editorconfig-checker): use the pre-packaged ec

WhoIsSethDaniel:


WhoIsSethDaniel:
This converts the `editorconfig-checker` install to use the pre-built binaries available via github.  `ec` creates *many* pre-built binaries for many different platforms. I chose the six that seemed most common within other mason installers. I can only test with linux and this installer WFM.

williamboman:
Thanks! I'll expand the platform detecting capabilities and add some more targets

PR #442, opened 2022-09-19
feat: add vscode-js-debug

Holi0317:


williamboman (commented):
Looks good! I wonder if we should change the name to `js-debug-adapter`? I've omitted `vscode` from package names before because I don't see a reason to include it

Holi0317:
Sure. I have update my branch.

Also renamed the binary name.

lua/mason-registry/js-debug-adapter/init.lua
```diff
@@ -0,0 +1,27 @@
...
+    ---@param ctx InstallContext
+    install = function(ctx)
+        local source = github.tag { repo = "microsoft/vscode-js-debug" }
+        source.with_receipt()
+        git.clone { "https://github.com/microsoft/vscode-js-debug", version = Optional.of(source.tag) }
+        ctx.spawn.npm { "install", "--legacy-peer-deps" }
+        ctx.spawn.npm { "run", "compile" }
+        ctx:link_bin(
+            "js-debug-adapter",
+            ctx:write_node_exec_wrapper("js-debug-adapter", path.concat { "out", "src", "debugServerMain.js" })
```
williamboman, line 24: I've been struggling to get things to work when executing the `debugServerMain.js` script. Using `vsDebugServer.js` however works fine. Are you using nvim-dap or something else?

Holi0317: Hmm interesting.

I am in fact using https://github.com/mxsdev/nvim-dap-vscode-js where it expect the path to the adapter's directory. So I just pass the mason package path to it.

Just read their code and seems they are also using `vsDebugServer.js`. I'll change the binary path then. Sorry about that.

williamboman: Cool! I've been playing around with that plugin recently as well. I'm planning on opening a PR that would allow passing an executable instead of the path to where the project is compiled, something like:

```
require("dap-vscode-js").setup({
    debugger_cmd = { "js-debug-adapter" }
})
```

williamboman (approved):
Awesome!

PR #472, opened 2022-09-27
shfmt now uses a pre-built package; no longer requires 'go' to build

WhoIsSethDaniel:


WhoIsSethDaniel:
tested on linux amd64. Unable to test other platforms.

williamboman (approved):
Great!

PR #477, opened 2022-09-28
fix(stylua): cannot be installed on arm64 Linux

wongjiahau:
This is because I'm using Neovim in Ubuntu that runs on Docker which runs on Apple M1 MacBook.


williamboman (approved):
:+1:

PR #546, opened 2022-10-12
Add prisma-fmt to the regestry

### LINKED ISSUE #393: [New package]: support `prisma-fmt`

- [ ] Yes

missing prisma-fmt to format prisma schemas

be able to format prisma schemas like prettier 

binary can be built with https://github.com/prisma/prisma-engines/tree/main/prisma-fmt

`prisma-fmt` setup for formatter.nvim
https://github.com/mhartington/formatter.nvim/commit/017b629241e7aa4e79280162f82766ddcb36f2fd

_No response_

### Additional context

_No response_

zhiburt:
Hi there

Close #393

Thanks for good first issue mark.

Take care

### DIFF
--- a/lua/mason-registry/index.lua
+++ b/lua/mason-registry/index.lua
@@ -150,6 +150,7 @@ return {
   ["powershell-editor-services"] = "mason-registry.powershell-editor-services",
   prettier = "mason-registry.prettier",
   prettierd = "mason-registry.prettierd",
+  ["prisma-fmt"] = "mason-registry.prisma-fmt",
   ["prisma-language-server"] = "mason-registry.prisma-language-server",
   proselint = "mason-registry.proselint",
   ["prosemd-lsp"] = "mason-registry.prosemd-lsp",
--- a/lua/mason-registry/prisma-fmt/init.lua
+++ b/lua/mason-registry/prisma-fmt/init.lua
@@ -1,0 +1,24 @@
+local Pkg = require "mason-core.package"
+local git = require "mason-core.managers.git"
+local platform = require "mason-core.platform"
+local path = require "mason-core.path"
+
+return Pkg.new {
+    name = "prisma-fmt",
+    desc = [[Prisma ORM formatter.]],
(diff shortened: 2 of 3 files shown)


williamboman (commented):
Cool! When the issue was created I took a quick look at it. After having done some research now again I've come to the same conclusion: The prisma formatter is generally bundled with the `prisma` npm package itself. This npm package ships with prebuilt binaries (which under the scenes are hosted by their "@prisma/engines" package). 

I'm a bit worried about landing this because:
1) It needs nightly rust toolchain to be installed. I'm pretty confident this means a vast majority will be excluded from using this package.
2) It uses an unstable source (git HEAD).

I'm wondering if it would be better to just provide a `prisma` package (which maps to the `prisma` npm package)? I also wonder why there's a `prisma-fmt` formatter in null-ls - exactly which executable is this formatter meant to target?

vuki656:
> Cool! When the issue was created I took a quick look at it. After having done some research now again I've come to the same conclusion: The prisma formatter is generally bundled with the `prisma` npm package itself. This npm package ships with prebuilt binaries (which behind the scenes are hosted by their "@prisma/engines" package).
> 
> I'm a bit worried about landing this because:
> 
> 1. It needs nightly rust toolchain to be installed. I'm pretty confident this means a vast majority will be excluded from using this package.
> 2. It uses an unstable source (git HEAD).
> 
> I'm wondering if it would be better to just provide a `prisma` package (which maps to the `prisma` npm package)? I also wonder why there's a `prisma-fmt` formatter in null-ls - exactly which executable is this formatter meant to target?

You're correct. It's bundled with Prisma and if you have Prisma lsp it should work as a language formatter. 

I think they extracted it to a separate package to make maintenance easier.

williamboman:
FWIW, they recommend installing Prisma as a project-local dependency: https://www.prisma.io/docs/concepts/components/prisma-cli/installation#local-installation-recommended. I'm not opposed to adding `prisma` as a standalone package in Mason though (PR welcome)

PR #565, opened 2022-10-17
feat(registry): add jsonlint tool for JSON

afranioce:


williamboman:
@williambotman /fixup

Issue #584, opened 2022-10-20
[New package]: nxls


nyarthan:
### I've searched open & closed issues for similar requests

Yes

### Package name

nxls

### Package homepage

https://github.com/nrwl/nx-console/tree/master/apps/nxls

### Languages

json

### How is this package distributed?

npm

Installation:
`npm install --location=global nxls`

williambotman:
Hello! Pull requests are always very welcomed to add new packages. If the distribution of the package is simple, the installation will most likely be so as well. See [CONTRIBUTING.md](https://github.com/williamboman/mason.nvim/blob/main/CONTRIBUTING.md) and the [API reference](https://github.com/williamboman/mason.nvim/blob/main/doc/reference.md) for more details! You may also use existing packages as reference.

<sup>` 🤖 This is an automated comment. `  [` 📖 Source code `](https://github.com/williamboman/botman)</sup>

PR #585, opened 2022-10-20
feat(registry): add nxls

nyarthan:
Closes: #584

### DIFF
--- a/lua/mason-registry/nxls/init.lua
+++ b/lua/mason-registry/nxls/init.lua
@@ -1,0 +1,13 @@
+local Pkg = require "mason-core.package"
+local npm = require "mason-core.managers.npm"
+
+return Pkg.new {
+    name = "nxls",
+    desc = [[A language server that utilizes the json-languageservice to provide code completion and more, for Nx workspaces.]],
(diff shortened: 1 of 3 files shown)


lua/mason-registry/nxls/init.lua
```diff
@@ -0,0 +1,13 @@
+local Pkg = require "mason-core.package"
+local npm = require "mason-core.managers.npm"
+
+return Pkg.new {
+    name = "nxls",
+    desc = [[A language server that utilizes the json-languageservice to provide code completion and more, for Nx workspaces.]],
```
williamboman, line 6: (I see this description is taken directly from their README). This is a bit confusing to me, does it depend on json-languageservice somehow? I'm assuming it bundles everything it needs in the npm package, and so this level of description might be a bit excessive and confusing. What about

```suggestion
    desc = [[A language server that provides code completion and more for Nx workspaces.]],
```

nyarthan: Yes, you'r right. It does bundle `vscode-json-languageservice`, so mentioning it is unnecessary. I use your suggestion.

williamboman: Cool, thanks for confirming!

PR #631, opened 2022-11-05
feat(registry): add autoflake

dirn:


williamboman:
@williambotman /fixup

PR #641, opened 2022-11-06
fix(protolint): use Protobuf language

williamboman:


williamboman:
I missed this in the original PR - @jayp0521 I feel like this might be more appropriate? For example the [buf](https://github.com/williamboman/mason.nvim/blob/b0439093e2d7365a06011d08490041942410afc2/PACKAGES.md#L672) CLI uses `Protobuf` as its language

PR #649, opened 2022-11-09
fix(rome): drop default version pin

otahontas:
Rome has been defaulting to quite an old version. They just released a new major 10.0.0, so I'd suggest dropping the old default and instead just going with the newest from npm.


williamboman (approved):
Great!

PR #650, opened 2022-11-09
feat(registry): add colorgen-nvim

mrjones2014:
Feel free to close if this is out of scope for `mason.nvim`.

I'd like to experiment with using `colorgen-nvim` to generate a colorscheme on-the-fly with a Neovim command in my configuration.

This PR adds `colorgen-nvim` as an installable package.


williamboman (approved):
Good enough overlap I'd say! `lemmy-help` is a very similar example

PR #676, opened 2022-11-19
Add nimpretty and nimsuggest

bolasm:


williamboman (changes requested):
Hey! It seems like both packages are missing their `install` property - without it the packages cannot be installed!

williamboman:
From what I can tell, these tools are bundled with the core distribution of Nim itself?

PR #706, opened 2022-11-29
chore: use pwsh instead of powershell

williamboman:
`powershell.exe` is apparently legacy software now.


OAguinagalde:
As far as I know, `powershell.exe` comes installed in every windows machine, while `pwsh` needs to be manually installed. It seems not even `windows11` comes with `pwsh` preinstalled https://www.technewstoday.com/how-to-update-powershell/, so this is basically an extra requirement for windows users. Is there a good reason for this change?

PR #720, opened 2022-12-03
fix: python health check problem in windows

zalgo3:
On my Windows 11, when I run `:healthcheck mason`, `:MasonLog` says the following:
```
Failed to spawn process. cmd="path/to/python3", err="ENOENT: no such file or directory"
```

Generally, the python executable on Windows is a .bat file, and I am guessing that the program cannot find it. So, I have added a conditional branch like `if platform.is.win then ... else ... end`. Could you review it?


williamboman (commented):
Thanks for sending a PR!

So I'm pretty confident python generally always ships as an `.exe` binary on Windows. What are you using to manage python?

The error message you provided (`Failed to spawn process. cmd="path/to/python3", err="ENOENT: no such file or directory"`) leads me to believe that you're probably running into an issue with `vim.g.python3_host_prog`. Is this global defined in your Neovim runtime? The "normal" python checks would not be producing an error message containing file path components.

Also, should the `.bat` vs `.exe` actually be an issue we'd have to revisit more things than just the healthcheck.

Finally, this is a libuv issue on Windows where it's unable to find `.cmd` and `.bat` scripts in `PATH`. This can pretty easily be worked around by resolving executables via Neovim APIs (`vim.fn.exepath()`) before sending it to `uv.spawn()`. It might be best to simply modify Mason's libuv process API to do this when spawning commands.

zalgo3:
@williamboman Thank you for reviewing! I use `pyenv` and my python is installed in `%USERPROFILE\.pyenv\pyenv-win\shims\python`. I feel you are right that it is not just a health check issue. The ideal solution would be to try the extensions registered in `%PATHEXT%`, as Windows actually does, but this may be a problem to be fixed on the `libuv` side. I found that alias is defined in `mason-core/spawn.lua` for programs that may have `.bat` or `.cmd`, so I suggest adding python there as well https://github.com/williamboman/mason.nvim/pull/720/commits/d912fa9b4c1f25dd84f65ba22b7adb219aac0455.

williamboman:
Cool! Unfortunately this will break for Windows users who don't use pyenv. We'll need to do this slightly differently. I've been wanting to get rid of that alias table and replace it with simply calling Neovim's `exepath()` (which fully respects PATHEXT) to expand the executable, let me provide an alternative solution

zalgo3:
@williamboman Thanks!

>I've been wanting to get rid of that alias table and replace it with simply calling Neovim's exepath() (which fully respects PATHEXT) to expand the executable, let me provide an alternative solution.

Is your solution like https://github.com/williamboman/mason.nvim/pull/720/commits/ee4f6d6b9176a0e5a073a316a0f3d73f44e9dbd1?

lua/mason-core/spawn.lua
```diff
@@ -85,12 +78,14 @@ setmetatable(spawn, {
                 }, cmd)
             end
 
-            local _, exit_code, signal = a.wait(function(resolve)
-                local handle, stdio, pid = process.spawn(cmd, spawn_args, resolve)
+            local _, exit_code, signal = a.wait(vim.schedule_wrap(function(resolve)
```
williamboman, line 81: Preferably we shouldn't have to schedule to Neovim's main loop, these functions should be able to execute outside of it. This is pretty much exactly what I was thinking though, will just have to think about a way to only schedule when there's a need to invoke exepath (there's an `is_executable` call some lines above that I think we can repurpose somehow)

zalgo3: I see. I'll give it some thought, but I'll hasten to report that `is_executable` actually returns true even for cmd that would cause `process.spawn` to fail.

zalgo3:
https://github.com/williamboman/mason.nvim/pull/773 fixes this PR.

PR #722, opened 2022-12-04
Update the name of the binary for editorconfig-checker

asmodeus812:
Changed the name of the binary for the editorconfig check linter to match what is expected by default from null ls. Currently, the name of the binary in the /bin folder does not match that.

### DIFF
--- a/lua/mason-registry/editorconfig-checker/init.lua
+++ b/lua/mason-registry/editorconfig-checker/init.lua
@@ -28,6 +28,6 @@ return Pkg.new {
         }
         source.with_receipt()
         local prog = source.asset_file:gsub("%.tar%.gz$", "")
-        ctx:link_bin("editorconfig-checker", path.concat { "bin", prog })
+        ctx:link_bin("ec", path.concat { "bin", prog })
     end,
 }


williamboman (commented):
Hey! This would be a breaking change. I actually looked into this when the package was added and I remember coming to the conclusion that `editorconfig-checker` is the accurate name for the binary (I believe Homebrew also ships it like that). Not sure where null-ls got the `ec` name from (which also is a pretty ambiguous two-letter name, generally reserved for core Unix utils).

You can change the default null-ls config during setup like so:

```lua
require("null-ls").setup {
  sources = {
      null_ls.builtins.diagnostics.editorconfig_checker.with { command = "editorconfig-checker" },
  }
}
```

Maybe [mason-null-ls.nvim](https://github.com/jay-babu/mason-null-ls.nvim) could help with translating these things.

asmodeus812:
> Hey! This would be a breaking change. I actually looked into this when the package was added and I remember coming to the conclusion that `editorconfig-checker` is the accurate name for the binary (I believe Homebrew also ships it like that). Not sure where null-ls got the `ec` name from (which also is a pretty ambiguous two-letter name, generally reserved for core Unix utils).
> 
> You can change the default null-ls config during setup like so:
> 
> ```lua
> require("null-ls").setup {
>   sources = {
>       null_ls.builtins.diagnostics.editorconfig_checker.with { command = "editorconfig-checker" },
>   }
> }
> ```
> 
> Maybe [mason-null-ls.nvim](https://github.com/jay-babu/mason-null-ls.nvim) could help with translating these things.

Hi, that is okay, i do agree that ec is ambigious, maybe i will report that to mason-null-ls to add as an exception to the default config. We can close this pr

jay-babu:
https://github.com/editorconfig-checker/editorconfig-checker/issues/232

the answer based on this discussion should resolve this discussion

PR #734, opened 2022-12-06
Add ripgrep to the list of runtime packages.

asmodeus812:
Adding a tool that is ubiquitous for it's usage in many nvim plugins, most notably telescope.

### DIFF
--- a/lua/mason-registry/ripgrep/init.lua
+++ b/lua/mason-registry/ripgrep/init.lua
@@ -1,0 +1,59 @@
+local Pkg = require "mason-core.package"
+local _ = require "mason-core.functional"
+local platform = require "mason-core.platform"
+local github = require "mason-core.managers.github"
+local path = require "mason-core.path"
+
+local coalesce, when = _.coalesce, _.when
+
+return Pkg.new {
+    name = "ripgrep",
+    desc = [[Ripgrep is a line-oriented search tool that recursively searches the current directory.]],
+    homepage = "https://github.com/BurntSushi/ripgrep",
+    languages = {},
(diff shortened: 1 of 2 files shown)


lua/mason-registry/ripgrep/init.lua
```diff
@@ -0,0 +1,59 @@
...
+                }
+                source.with_receipt()
+                ctx.fs:rename(source.asset_file:gsub("%.zip$", ""), "ripgrep")
+                ctx:link_bin("rg", path.concat { "ripgrep" , "rg.exe" }) -- link binary
+            end,
+            linux = function()
+                local source = github.untargz_release_file {
+                    repo = repo,
+                    asset_file = coalesce(
+                        -- @Note: using the x86 build instead of arm which is not available yet
```
asmodeus812, line 38: Should work fine with the m1 translation layer for now, i have been using it on an m1 for a while.

williamboman (commented):
Hey! Thanks for the PR! I'm of the opinion that `rg` has too small of an overall overlap with Neovim to justify adding it to Mason, for the time being. It's a very versatile tool that has many use cases outside of Neovim. I do understand this would be helpful for some people as some popular Neovim plugins makes use of `rg`, but adding `rg` would be a bit like adding `git` to Mason; a bit out of scope for now!

williamboman:
And as a side-note: I'm reworking how the registry works and will soon enable adding custom registries. This would allow people to manager whatever packages they want to under Mason!

williamboman:
Cool, yeah that'll do it :+1:! For posterity: do note that doing this is not officially supported and may likely break some time in the future.

PR #757, opened 2022-12-12
feat(semgrep): onboard semgrep

jay-babu:
Resolves: https://github.com/jay-babu/mason-null-ls.nvim/issues/38#issuecomment-1342832050


lua/mason-registry/semgrep/init.lua
```diff
@@ -0,0 +1,18 @@
...
+  name = "semgrep",
+  desc = "Semgrep is a fast, open-source, static analysis engine for finding bugs, detecting vulnerabilities in third-party dependencies, and enforcing code standards.",
+  homepage = "https://github.com/returntocorp/semgrep",
+  languages = {
+    Pkg.Lang.Go,
+    Pkg.Lang.Java,
+    Pkg.Lang.Python,
+    Pkg.Lang.Ruby,
+    Pkg.Lang.TypeScript,
+    Pkg.Lang.TypeScriptReact,
```
jay-babu, line 14: this is new, please let me know if you would like to change

williamboman: Hm I think it might be too specific. These language entries are really just meant to provide the means to do cursory lookups, as an entrypoint for further discovery.

lua/mason-registry/semgrep/init.lua
```diff
@@ -0,0 +1,18 @@
...
+    name = "semgrep",
+    desc = "Semgrep is a fast, open-source, static analysis engine for finding bugs, detecting vulnerabilities in third-party dependencies, and enforcing code standards.",
+    homepage = "https://github.com/returntocorp/semgrep",
+    languages = {
+        Pkg.Lang.Go,
+        Pkg.Lang.Java,
+        Pkg.Lang.Python,
+        Pkg.Lang.Ruby,
+        Pkg.Lang.TypeScript,
+        Pkg.Lang.TypeScriptReact,
```
williamboman, line 14: I found some more languages in their README 🕵️!
```suggestion
        Pkg.Lang["C#"],
        Pkg.Lang.Go,
        Pkg.Lang.Java,
        Pkg.Lang.JavaScript,
        Pkg.Lang.JSON,
        Pkg.Lang.PHP,
        Pkg.Lang.Python,
        Pkg.Lang.Ruby,
        Pkg.Lang.TypeScript,
        Pkg.Lang.Scala,
```

jay-babu: Oh, thanks. Much appreciated!

williamboman:
@williambotman /fixup

lua/mason-registry/semgrep/init.lua
williamboman: I see! What I'm worried about is over-saturating the language list with all kinds of language variants and flavours - it might actually end up negatively impact discoverability if there are too many options.

jay-babu: That's fine, I have no qualms either way

PR #773, opened 2022-12-19
fix(spawn): always expand cmd if PATH is not modified

williamboman:
Closes #720.


zalgo3 (approved):
It seems to work ultimately! Nice job!

PR #777, opened 2022-12-20
refactor(registry): remove xo

thanhvule0310:
Since version [v0.41.0](https://github.com/xojs/xo/releases/tag/v0.41.0), XO can no longer be installed globally.


williamboman:
Thanks! Will have to do some work on dealing with removed packages. Removing the package definition will simply make the package disappear entirely from all Mason interfaces, it should preferably still surface as an installed package if it's installed (enough information is persisted on fs already so this is just a matter of implementation).

williamboman:
Closing this as packages are no longer added to this registry, see instead https://github.com/mason-org/mason-registry/.

PR #782, opened 2022-12-21
fix(powershell): use pwsh if available

williamboman:


williamboman:
@williambotman /fixup

PR #792, opened 2022-12-23
feat(dotnet/init.lua): ignore failed sources

stonelasley:
I contract for a number of organizations and I'm never authenticated to all the private nuget feeds at once so installing tools always fails without this flag. I believe this should be a safe addition.

[Docs](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-tool-install)

`--ignore-failed-sources`

Treat package source failures as warnings.

### DIFF
--- a/lua/mason-core/managers/dotnet/init.lua
+++ b/lua/mason-core/managers/dotnet/init.lua
@@ -31,6 +31,7 @@ function M.install(pkg, opt)
     ctx.spawn.dotnet {
         "tool",
         "update",
+        "--ignore-failed-sources",
         "--tool-path",
         ".",
         ctx.requested_version
--- a/tests/mason-core/managers/dotnet_spec.lua
+++ b/tests/mason-core/managers/dotnet_spec.lua
@@ -12,6 +12,7 @@ describe("dotnet manager", function()
             assert.spy(ctx.spawn.dotnet).was_called_with {
                 "tool",
                 "update",
+                "--ignore-failed-sources",
                 "--tool-path",
                 ".",
                 { "--version", "42.13.37" },


williamboman:
Hello! I'm struggling a bit to understand what "package source failure" means?

stonelasley:
> Hello! I'm struggling a bit to understand what "package source failure" means?

William, it is a bit confusing. A source could fail for a number of reasons, either you aren't authenticated, or the source can't be reached e.g. it's only available while you're on a VPN. By default, if you install a tool and any of the sources aren't reachable you can't install even if the package is from the api.nuget.org source which is available.  Adding the ignore flag doesn't require all your sources to be reachable or for you to be authenticated to them all, as long as the source that provides the package is available it will install. 

```xml1   <?xml version="1.0" encoding="utf-8"?>
  1<configuration>
     <packageSources>
       <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" /> // this is the default `source`
       <add key="MyPrivateSource" value="https://pkgs.dev.azure.com/someorg/_packaging/MyPrivateSource/nuget/v3/index.json" /> // this is a source that requires authentication
       <add key="MyInternalSource" value="10.100.0.1/nuget/v3/index.json" /> // this is a source that's only reachable while on VPN
     </packageSources>
   </configuration>
`
Say my configuration looked like the above. To use mason to install csharpier I would have to be authenticated to the MyPrivateSource and be on a network where MyInternal is reachable, even though I only really need api.nuget.org to serve the package I'm installing. 

An argument against this PR would be that you ideally shouldn't have private sources in your root nuget.config as they should be configured within the project. My only counter to that is working with organizations that have a spectrum of maturity in their environments makes this tricky.

**Comment by williamboman** [comment]
@williambotman /apply

```diff
diff --git a/tests/mason-core/managers/dotnet_spec.lua b/tests/mason-core/managers/dotnet_spec.lua
index ffcbd33..b633212 100644
--- a/tests/mason-core/managers/dotnet_spec.lua
+++ b/tests/mason-core/managers/dotnet_spec.lua
@@ -12,6 +12,7 @@ describe("dotnet manager", function()
             assert.spy(ctx.spawn.dotnet).was_called_with {
                 "tool",
                 "update",
+                "--ignore-failed-sources",
                 "--tool-path",
                 ".",
                 { "--version", "42.13.37" },
```

williamboman (approved):
Ah thanks for the explanation, makes sense!

PR #809, opened 2022-12-26
chore(CI): add more runners for tests workflow

williamboman:


williamboman:
Not sure why this is failing and decided I for the time being don't care, will revisit

Issue #392, opened 2022-09-07
Sub-packages


williamboman:
### I've searched open issues for similar requests

- [X] Yes

### Is your feature request related to a problem? Please describe.

placeholder

### Describe the solution you'd like

placeholder

znd4:
workaround snippet (requires [plenary.nvim](https://github.com/nvim-lua/plenary.nvim)):

```lua
local function mason_package_path(package)
    local path = vim.fn.resolve(vim.fn.stdpath("data") .. "/mason/packages/" .. package)
    return path
end

-- depends on package manager / language
local command = "./venv/bin/pip"
local args = { "install", "pylsp-rope", "python-lsp-ruff", "pyls-isort", "python-lsp-black" }

require("plenary.job")
    :new({
        command = command,
        args = args,
        cwd = mason_package_path("python-lsp-server"),
    })
    :start()
```

shubham-cpp:
Hello, I would to add [styled components](https://github.com/styled-components/typescript-styled-plugin) should I follow the same method in #166  [Prettierd plugins](https://github.com/williamboman/mason.nvim/issues/161)

PR #822, opened 2022-12-28
feat(registry): add python-lsp-ruff

forrestfwilliams:
Enable installation of the python-lsp-server ruff extension.


williamboman (commented):
Nice! It seems like this is a plugin to python-lsp, and requires to be installed in the same virtual environment as python-lsp itself. This is not yet supported in mason (see #392). However, `mason-lspconfig` provides a `:PylspInstall` command for this exact purpose. Let's add it there instead!

forrestfwilliams:
Sounds good!

PR #853, opened 2023-01-05
add Markdown JSX Linter: eslint-mdx

Cybergenik:
Basically, just looked at the `alex` and `eslint-lsp` Packages for reference on integration.


williamboman:
@williambotman /fixup

williamboman:
Hey! This seems to be a plugin to ESLint, and not a standalone linter? I'm not sure it'd make sense to install it as a separate Mason package, I also noticed it doesn't provide a `eslint-plugin-mdx` executable so the `bin` is a bit problematic.

williamboman:
https://github.com/williamboman/mason.nvim/issues/392 aims to enable installing "sub-packages", I feel like this would fit under that category, no?

Cybergenik:
Hey, Thanks for maintaining Mason!

> #392 aims to enable installing "sub-packages", I feel like this would fit under that category, no?

Yeah, I think that makes sense. As it's more of a plugin for the eslint lsp.

williamboman:
Ok, let's close this for now then!

PR #868, opened 2023-01-08
feat(registry): add fennel-ls

### LINKED ISSUE #867: [New package]: fennel language server
### I've searched open & closed issues for similar requests

Yes

### Package name

fennel-ls

### Package homepage

https://sr.ht/~xerool/fennel-ls/

### Languages

fennel

The package is available through sourcehut using git. It has to be cloned, and then `make` has to be run.

Jeansidharta:
Closes #867

This PR adds support for the fennel language server.

It's repo is [here](https://sr.ht/~xerool/fennel-ls)
And `nvim-lspconfig`'s entry on it is [here](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#fennel-ls)

One problem with this implementation is that it relies on the presence of `make` in the user's system. This is probably a safe bet on unix systems, but not on windows systems. Are there any suggestions on how to handle this?

### DIFF
--- a/PACKAGES.md
+++ b/PACKAGES.md
@@ -76,6 +76,7 @@
 - [eslint-lsp](#eslint-lsp)
 - [eslint_d](#eslint_d)
 - [fantomas](#fantomas)
+- [fennel-ls](#fennel-ls)
 - [firefox-debug-adapter](#firefox-debug-adapter)
 - [fixjson](#fixjson)
 - [flake8](#flake8)
@@ -1760,6 +1761,18 @@ Categories: `Formatter`
 
 ```
 :MasonInstall fantomas
+```
+
+# fennel-ls
+
+> fennel-ls is a language server for the fennel language
+
+Homepage: [https://sr.ht/~xerool/fennel-ls](https://sr.ht/~xerool/fennel-ls)  
+Languages: `Fennel`  
+Categories: `LSP`  
+
+```
+:MasonInstall fennel-ls
 ```
 
 
--- a/lua/mason-registry/fennel-ls/init.lua
+++ b/lua/mason-registry/fennel-ls/init.lua
@@ -1,0 +1,17 @@
+local Pkg = require "mason-core.package"
+local _ = require "mason-core.functional"
+local git = require "mason-core.managers.git"
+
+return Pkg.new {
+    name = "fennel-ls",
+    desc = [[Fennel language server]],
+    languages = { Pkg.Lang.Fennel },
+    categories = { Pkg.Cat.LSP },
+    homepage = "https://sr.ht/~xerool/fennel-ls",
+    ---@async
+    ---@param ctx InstallContext
+    install = function(ctx)
+        git.clone({ "https://git.sr.ht/~xerool/fennel-ls" }).with_receipt()
+        ctx.spawn.make {}
+    end,
+}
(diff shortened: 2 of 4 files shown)


Jeansidharta:
> > One problem with this implementation is that it relies on the presence of make in the user's system. This is probably a safe bet on unix systems, but not on windows systems. Are there any suggestions on how to handle this?
> 
> Do you think they'd be open to publishing this package on Luarocks?

I have no idea, but seems like a good solution. I just opened [this](https://todo.sr.ht/~xerool/fennel-ls/1) issue on the repo, and will wait for a response

Jeansidharta:
Since #971 was merged a few months ago, and the owner of fennel-ls has not responded, I'll be closing this PR for now

XeroOl:
Apologies for not responding earlier. I am the owner of fennel-ls.

> Do you think they'd be open to publishing this package on Luarocks?

I looked into making a LuaRocks package for fennel-ls, but it doesn't seem like it would help with building on Windows.
* if I use `type="builtin"`, there's no way to compile fennel-ls
* if I use `type="make"`, we can make a LuaRocks package for fennel-ls, but it doesn't solve the Windows problem because it depends on make.
* if I use `type="command"`, there's no way to do specify the alternate command for building on Windows, and apparently you can't set environment variables the same way on Windows and unix.
* if I create a custom type of rock (which I did, at [luarocks-build-fennel](https://luarocks.org/modules/xerool/luarocks-build-fennel)), LuaRocks fails to install if you're using `--tree` (which mason will want to do, as we don't want to touch the system files), because LuaRocks only loads the build types from the system's rocks.

So, I've given up on LuaRocks as a way to make fennel-ls build on windows.
I'm not even sure if it can run on Windows at all. Fennel-LS is a Lua script with `#!/bin/env lua` at the top, and I don't think Windows will figure out the shebang unless you changed the launch command to `lua.exe /path/to/fennel-ls`.

Some alternative ways to build:
1. Download the prebuilt script from https://git.sr.ht/~xerool/fennel-ls/refs/download/0.1.0/fennel-ls
   Unfortunately, it seems like mason would need to update the URL each time a release comes out. I would be happy to send a PR in every time.
2. Custom build command.
   Instead of triggering `make`, we can hardcode the command `lua ./fennel --add-package-path "./src/?.lua" --add-fennel-path "./src/?.fnl" --require-as-include --compile src/fennel-ls.fnl`. This is the command that `make` would have run anyway. I would be happy to send in a PR every time this changes, and I believe it will be able to build fennel-ls for the forseeable future, but I also understand if this isn't desired.
3. Decide that fennel-ls works unix only.
   I'm not sure if there's precedent for this in mason.nvim, but I believe it may be okay to skip Windows support in this case. In the 2023 Fennel Survey, only 6 people claimed to use fennel on Non-WSL Windows, out of 136 responses, so I don't think there's much demand or expectation for fennel tools to work well with Windows.

XeroOl:
Good news! LuaRocks seems to have fixed their tree issue, so fennel-ls can work on LuaRocks!   🎉
https://github.com/mason-org/mason-registry/pull/4648

garrett-hopper:
@XeroOl, any idea when a new version might be released and made available in luarocks? (Specifically looking at the `fnlsproject.fnl` config stuff 🙂)

XeroOl:
@garrett-hopper i did it 20 days ago and they haven't updated yet

PR #871, opened 2023-01-09
fix(editorconfig-checker): use ec instead of editorconfig-checker for binary

jemag:
Fixes https://github.com/williamboman/mason.nvim/issues/870


williamboman:
https://github.com/williamboman/mason.nvim/issues/870#issuecomment-1432597542

PR #887, opened 2023-01-12
feat: add buffer name

fidgetingbits:
This is a simple change to make the buffer used by mason have a specific name.

The reason for this change is that I use a voice coding system that relies on reading out window titles to establish a specific context to enable and disable certain command grammars, and for plugin specific windows the easiest way to differentiate is to use the buffer name which can be placed into the vim title.

### DIFF
--- a/lua/mason-core/ui/display.lua
+++ b/lua/mason-core/ui/display.lua
@@ -394,6 +394,7 @@ function M.new_view_only_win(name, filetype)
         for key, value in pairs(buf_opts) do
             vim.api.nvim_buf_set_option(bufnr, key, value)
         end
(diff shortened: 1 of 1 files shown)


fidgetingbits:
worth noting folke's comment here https://github.com/folke/lazy.nvim/pull/378, so this might not be something you can easily merge either

williamboman:
Yeah that's what I was thinking as well, was double checking it right now. Can't you leverage the `filetype` autocmd?

fidgetingbits:
Ya that will work actually, I missed that most plugins are setting filetype to something unique, but I can put that into the titlestring which I can then pull out with the voice system. Thanks for the suggestion.

PR #897, opened 2023-01-13
feat: use switch to checkout tags

Decodetalkers:
Log: use switch


Decodetalkers:
Feel so sorry.. but I think we should use switch , so may be should change to draft?

PR #898, opened 2023-01-13
feat(registry): add drools-lsp

errantepiphany:
This PR adds support for Red Hat's [Drools Rule Language](https://docs.drools.org/latest/drools-docs/docs-website/drools/language-reference/index.html#con-drl_drl-rules) (ft=`drools`, ext=`.drl`). The [drools-lsp](https://github.com/kiegroup/drools-lsp/) server config was recently merged into [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) (also written by me). The generated doc link in nvim-lspconfig is [here](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#drools_lsp).

The companion PR in mason-lspconfig.nvim is [here](https://github.com/williamboman/mason-lspconfig.nvim/pull/125).

_Thank you so much!_

Signed-off-by: David Ward <dward@redhat.com>

### DIFF
--- a/lua/mason-registry/drools-lsp/init.lua
+++ b/lua/mason-registry/drools-lsp/init.lua
@@ -1,0 +1,71 @@
+--[[
+Configuration:
+
+-- You can use the Mason-generated shell execution wrapper:
+require("lspconfig").drools_lsp.setup {
+    capabilities = capabilities,
+    on_attach = on_attach,
+    cmd = { 'drools-lsp' },
+}
+
+-- or you can specify the path to the Mason-installed drools jar file (allowing you to also specify java settings):
+local util = require "lspconfig.util"
+local reg = require "mason-registry"
+local jar = "drools-lsp-server-jar-with-dependencies.jar"
+require("lspconfig").drools_lsp.setup {
+    capabilities = capabilities,
+    on_attach = on_attach,
+    settings = {
+        -- optional
+        java = {
+            bin = "/path/to/java",
+            opts = { "-Xmx500m" },
+        },
+        -- required
+        drools = {
+            jar = util.path.join(reg.get_package("drools-lsp"):get_install_path(), jar),
+            -- the above is *currently* equivalent to the below (but path below could change in the future)
+            -- jar = vim.fn.stdpath "data" .. "/mason/packages/drools-lsp/" .. jar,
+        },
+    },
+}
+
+-- or you can use any of the other configuration options listed here:
+https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#drools_lsp
+--]]
+
+local Pkg = require "mason-core.package"
+local github = require "mason-core.managers.github"
+local Optional = require "mason-core.optional"
+local path = require "mason-core.path"
+
+local name = "drools-lsp"
+local repo = "kiegroup/" .. name
+local file = name .. "-server-jar-with-dependencies.jar"
+
+return Pkg.new {
+    name = name,
+    desc = [[An implementation of a language server for the Drools Rule Language.]],
+    homepage = "https://github.com/" .. repo,
+    languages = { Pkg.Lang.Drools },
+    categories = { Pkg.Cat.LSP },
+    ---@async
+    ---@param ctx InstallContext
+    install = function(ctx)
+        github
+            .download_release_file({
+                repo = repo,
+                version = Optional.of "latest",
+                asset_file = file,
+                out_file = file,
+            })
+            .with_receipt()
+        ctx:link_bin(
+            name,
+            ctx:write_shell_exec_wrapper(
+                name,
+                ("java -cp %q org.drools.lsp.server.Main"):format(path.concat { ctx.package:get_install_path(), file })
+            )
+        )
+    end,
+}
--- a/lua/mason-registry/drools-lsp/init.lua
+++ b/lua/mason-registry/drools-lsp/init.lua
@@ -1,70 +1,31 @@
---[[
-Configuration:
-
--- You can use the Mason-generated shell execution wrapper:
-require("lspconfig").drools_lsp.setup {
-    capabilities = capabilities,
-    on_attach = on_attach,
-    cmd = { 'drools-lsp' },
-}
-
--- or you can specify the path to the Mason-installed drools jar file (allowing you to also specify java settings):
-local util = require "lspconfig.util"
-local reg = require "mason-registry"
-local jar = "drools-lsp-server-jar-with-dependencies.jar"
-require("lspconfig").drools_lsp.setup {
-    capabilities = capabilities,
-    on_attach = on_attach,
-    settings = {
-        -- optional
-        java = {
-            bin = "/path/to/java",
-            opts = { "-Xmx500m" },
-        },
-        -- required
-        drools = {
-            jar = util.path.join(reg.get_package("drools-lsp"):get_install_path(), jar),
-            -- the above is *currently* equivalent to the below (but path below could change in the future)
-            -- jar = vim.fn.stdpath "data" .. "/mason/packages/drools-lsp/" .. jar,
-        },
-    },
-}
-
--- or you can use any of the other configuration options listed here:
-https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#drools_lsp
---]]
-
 local Pkg = require "mason-core.package"
 local github = require "mason-core.managers.github"
 local Optional = require "mason-core.optional"
 local path = require "mason-core.path"
 
-local name = "drools-lsp"
-local repo = "kiegroup/" .. name
-local file = name .. "-server-jar-with-dependencies.jar"
-
 return Pkg.new {
-    name = name,
+    name = "drools-lsp",
     desc = [[An implementation of a language server for the Drools Rule Language.]],
-    homepage = "https://github.com/" .. repo,
+    homepage = "https://github.com/kiegroup/drools-lsp",
     languages = { Pkg.Lang.Drools },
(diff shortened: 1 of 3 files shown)


errantepiphany:
I recognize that having a separate file (`drools_lsp/util.lua`) in addition to the standard init file (`drools_lsp/init.lua`) is "new", since no other server in the registry does that. But it's the best way I could think of for a user to easily configure the path to the installed jar file. Here are all the options I could think of, in decreasing preference order:
```lua
-- strategy 1: util.lua
require('lspconfig').drools_lsp.setup {
  settings = {
    drools = {
      jar = require('mason-registry.drools-lsp.util').get_drools_jar(),
    },
  },
}

-- strategy 2: mason-core.path
require('lspconfig').drools_lsp.setup {
  settings = {
    drools = {
      jar = (function()
        local path = require 'mason-core.path'
        local pkg = path.package_prefix 'drools-lsp'
        return path.concat { pkg, 'drools-lsp-server-jar-with-dependencies.jar' }
      end)(),
    },
  },
}

-- strategy 3: mason-core.path and vim.fn.stdpath
require('lspconfig').drools_lsp.setup {
  settings = {
    drools = {
      jar = (function()
        local path = require 'mason-core.path'
        local pkg = path.concat { vim.fn.stdpath 'data', 'mason', 'packages', 'drools-lsp' }
        return path.concat { pkg, 'drools-lsp-server-jar-with-dependencies.jar' }
      end)(),
    },
  },
}

-- strategy 4: hardcoded path
require('lspconfig').drools_lsp.setup {
  settings = {
    drools = {
      jar = vim.fn.expand '$HOME/.local/share/nvim/mason/packages/drools-lsp/drools-lsp-server-jar-with-dependencies.jar',
    },
  },
}
```
I thought this would be okay, considering each server has its own directory in the registry, so the util.lua file is "scoped", so to speak, for that server. If only ever having one lua file for each server in the registry is the rule, then why have a separate directory per server? Instead of `myserver/init.lua`, it would just be `myserver.lua`. But having directories makes me feel like it was an intentional choice to enable this type of thing. I'm actually kind of surprised that no other server has leveraged this. Then again, maybe I'm missing something fundamental here. :)

errantepiphany:
I just came up with a fourth option (and added it to the code):
```lua
-- option 5: shell exec wrapper
-- ~/.config/nvim/init.lua
require('lspconfig').drools_lsp.setup({
    cmd = { 'drools-lsp' },
})
-- mason.nvim/lua/mason-registry/drools-lsp/init.lua
-- ...
return Pkg.new({
    -- ...
        ctx:link_bin(
            "drools-lsp",
            ctx:write_shell_exec_wrapper(
                "drools-lsp",
                ("java -cp %q org.drools.lsp.server.Main"):format(path.concat {
                    ctx.package:get_install_path(),
                    "drools-lsp-server-jar-with-dependencies.jar",
                })
            )
        )
    -- ...
})
```
Would you still be okay with me leaving the util.lua? I ask because the downside of the option above is that it's not allowing the user to specify the java path or options. I could make the `drools-lsp` script more complicated so that it can receive and process arguments and then turn them into options to put into the java command. But I don't see other servers in the registry that do that... Thougts?

errantepiphany:
@williamboman Okay, I've removed util.lua, and now only init.lua remains. As for configuration options, please look at the comment section in the top of init.lua. Does all that look better to you now? (I've tested both options.) Thanks!

williamboman:
Great! Yeah both option 4 & 5 would make sense. Not sure if you're involved in the project/community, but do you think they'd be open to adding a manifest to the JAR file? Specifically thinking about providing a `Main-Class` for the application's entrypoint. This would make porting it to https://github.com/mason-org/mason-registry a bit simpler as it currently is unable to express a `java -cp <path> <mainclass>` executable wrapper.

williamboman:
@williambotman /apply

```diff
diff --git a/lua/mason-registry/drools-lsp/init.lua b/lua/mason-registry/drools-lsp/init.lua
index 54e92f6..e841afc 100644
--- a/lua/mason-registry/drools-lsp/init.lua
+++ b/lua/mason-registry/drools-lsp/init.lua
@@ -1,70 +1,31 @@
---[[
-Configuration:
-
--- You can use the Mason-generated shell execution wrapper:
-require("lspconfig").drools_lsp.setup {
-    capabilities = capabilities,
-    on_attach = on_attach,
-    cmd = { 'drools-lsp' },
-}
-
--- or you can specify the path to the Mason-installed drools jar file (allowing you to also specify java settings):
-local util = require "lspconfig.util"
-local reg = require "mason-registry"
-local jar = "drools-lsp-server-jar-with-dependencies.jar"
-require("lspconfig").drools_lsp.setup {
-    capabilities = capabilities,
-    on_attach = on_attach,
-    settings = {
-        -- optional
-        java = {
-            bin = "/path/to/java",
-            opts = { "-Xmx500m" },
-        },
-        -- required
-        drools = {
-            jar = util.path.join(reg.get_package("drools-lsp"):get_install_path(), jar),
-            -- the above is *currently* equivalent to the below (but path below could change in the future)
-            -- jar = vim.fn.stdpath "data" .. "/mason/packages/drools-lsp/" .. jar,
-        },
-    },
-}
-
--- or you can use any of the other configuration options listed here:
-https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#drools_lsp
---]]
-
 local Pkg = require "mason-core.package"
 local github = require "mason-core.managers.github"
 local Optional = require "mason-core.optional"
 local path = require "mason-core.path"
 
-local name = "drools-lsp"
-local repo = "kiegroup/" .. name
-local file = name .. "-server-jar-with-dependencies.jar"
-
 return Pkg.new {
-    name = name,
+    name = "drools-lsp",
     desc = [[An implementation of a language server for the Drools Rule Language.]],
-    homepage = "https://github.com/" .. repo,
+    homepage = "https://github.com/kiegroup/drools-lsp",
     languages = { Pkg.Lang.Drools },
     categories = { Pkg.Cat.LSP },
     ---@async
     ---@param ctx InstallContext
     install = function(ctx)
+        local jar = "drools-lsp-server-jar-with-dependencies.jar"
         github
             .download_release_file({
-                repo = repo,
+                repo = "kiegroup/drools-lsp",
                 version = Optional.of "latest",
-                asset_file = file,
-                out_file = file,
+                asset_file = jar,
+                out_file = jar,
             })
             .with_receipt()
         ctx:link_bin(
-            name,
+            "drools-lsp",
             ctx:write_shell_exec_wrapper(
-                name,
-                ("java -cp %q org.drools.lsp.server.Main"):format(path.concat { ctx.package:get_install_path(), file })
+                "drools-lsp",
+                ("java -cp %q org.drools.lsp.server.Main"):format(path.concat { ctx.package:get_install_path(), jar })
             )
         )
     end,
```

lua/mason-registry/drools-lsp/init.lua
```diff
@@ -0,0 +1,71 @@
+--[[
```
williamboman: This documentation is great, but it's not the best place to put it. I see there's a great deal of documentation provided in the lspconfig server config, hopefully I didn't remove anything that isn't there already?

williamboman (approved):
I applied some minor stylistic changes to make things a bit more grep-friendly. Thanks!!

lua/mason-registry/drools-lsp/init.lua
errantepiphany: @williamboman It looks all good. I'll switch over my local testing nvim init.lua to use the public github location vs. my local development folders now, and re-test. But again, it looks like the changes are fine. As for the removed documentation at the top of the file, the only thing of real loss was the option of simply configuring the cmd to the exec-shell-wrapper that mason generates. That part is only available to mason users. But that's fine too. I'm going to put more documentation right into the README on the drools-lsp github repo. :)

williamboman: > That part is only available to mason users. But that's fine too. I'm going to put more documentation right into the README on the drools-lsp github repo. :)

Ah, we should be able to provide any necessary extra configuration in mason-lspconfig. [This directory](https://github.com/williamboman/mason-lspconfig.nvim/tree/main/lua/mason-lspconfig/server_configurations) contains configuration overrides provided by Mason.

errantepiphany: Ah, something like [this](https://github.com/williamboman/mason-lspconfig.nvim/blob/main/lua/mason-lspconfig/server_configurations/groovyls/init.lua) seems like it worked. But how does one leverage that mechanism? Or is it just automatic? I'll create a new branch, make the change, test it, and if all good, submit a new PR for that.

Oh, and btw, I re-tested with your changes straight from github main, and it all works still. So your changes were fine. :)

williamboman: Awesome, thanks!! [Yep it's automatic.](https://github.com/williamboman/mason-lspconfig.nvim/blob/43c7f402cb9822b61b5dc05bf0b385928df256a6/lua/mason-lspconfig/lspconfig_hook.lua#L26) I saw you had a question regarding the directory structure and it's the same structure over there - directories instead of files. It's just a personal preference of mine (in the case of mason-lspconfig it does allow for additional, adjacent, modules should they be needed).

The directory name needs to match up with the lspconfig server name.

errantepiphany: Although oops, I must have missed putting drools-lsp in https://github.com/williamboman/mason.nvim/blob/main/PACKAGES.md, though... But I think you do regular re-generates, yes?

errantepiphany:
I'm actually the Engineering Manager for the Drools team. ;) So yes, I can have it added.

PR #920, opened 2023-01-17
feat(registry): Add terraform-docs

### LINKED ISSUE #826: [New package]: terraform-docs
### I've searched open & closed issues for similar requests

Yes

### Package name

terraform-docs

### Package homepage

https://github.com/terraform-docs/terraform-docs/

### Languages

terraform

GitHub releases

hermitmaster:
Closes #826

### DIFF
--- a/lua/mason-registry/index.lua
+++ b/lua/mason-registry/index.lua
@@ -237,6 +237,7 @@ return {
   taplo = "mason-registry.taplo",
   ["teal-language-server"] = "mason-registry.teal-language-server",
   tectonic = "mason-registry.tectonic",
+  ["terraform-docs"] = "mason-registry.terraform-docs",
   ["terraform-ls"] = "mason-registry.terraform-ls",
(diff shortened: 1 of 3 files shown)


williamboman:
Hey! Hm what relation does terraform-docs have with Neovim? For the time being there has to be some kind of broader connection to Neovim to justify supporting a package

hermitmaster:
It’s something you’d likely wire up with null-ls to automatically update docs on write.

williamboman:
Ah I see. To me it sounds like this would be out of scope for Mason currently. The categories (LSP, DAP, Formatter, Linter) serves as a pretty good guide for what packages are in scope. There are some packages that don't fit within these categories, but these have a demonstrable connection to Neovim, and I'm afraid I don't quite see it here. The good news is that custom registries will soon:tm: be supported, which would allow people to extend the core registry with whatever tools they wish to manage through Mason.

PR #925, opened 2023-01-18
fix(drools-lsp): remove unused shell exec wrapper

errantepiphany:
fix(drools-lsp): remove unused shell exec wrapper

Signed-off-by: David Ward <dward@redhat.com>


errantepiphany:
Thanks!

PR #938, opened 2023-01-22
add armv71 alias for arm64 arch

bainos:


bainos:
Tested in Termux on Lenovo Tab M7

```bash
$ uname -a
Linux localhost 4.19.127-g77c031a-dirty #1 SMP PREEMPT Fri Oct 21 16:18:30 CST 2022 armv7l Android
```

Mason null-ls config

```
"luacheck", 
"stylua",
"black", 
"flake8",
"shellcheck", 
"shfmt", 
"yamlfmt", 
 --"jq", -- masom fails for arm64 arch
-- markdown 
"markdownlint" ,
```

williamboman (commented):
Hey! Hm what leads you to want to interpret armv7l (32 bit) as compatible with arm64?

bainos:
You are right. I should have chalked it before pushing a PR.

Sorry.

I tested 'arm32' and this let Mason work a little bit better, at least downloading some providers instead of failing by default.

I'll try to go deep if I can, because by now I cannot understand why, for example, Mason fails downloading stylua, which I can installa via pkg/apt.

Any suggestion is welcome.

williamboman:
I see! So `arm32` is not a recognized architecture in Mason. The reason installation was failing was probably because the package simply isn't supported on armv7/6 architectures. Few packages currently are.

PR #945, opened 2023-01-26
feat(registry): add veryl language server

dalance:
I want to add veryl language server.


williamboman (approved):
Cool! I ported this to https://github.com/mason-org/mason-registry/pull/356 which will soon:tm: be the new registry source

PR #957, opened 2023-02-01
feat(jdtls): include a decompiler (quiltflower) with jdtls

jay-babu:
This can be a separate package, but the jar is very small `997k`


williamboman (commented):
How is this decompiler generally used in the context of Neovim/jdtls? I can't find any mentions of it in eclipde.jdtls or nvim-jdtls

jay-babu:
fernflower is what people commonly use. It is used to decompile library code. There aren't many references to it, but it is used widely. Essentially to look at the source code from a compiled jar or attempt to do so, since it's a decompiler.

PR #958, opened 2023-02-01
feat(registry): add google-java-format

jay-babu:


jay-babu:
https://github.com/williamboman/mason.nvim/issues/951

lua/mason-registry/google-java-format/init.lua
```diff
@@ -0,0 +1,33 @@
...
+    ---@async
+    ---@param ctx InstallContext
+    install = function(ctx)
+        github
+            .download_release_file({
+                repo = "google/google-java-format",
+                out_file = "google-java-format.jar",
+                asset_file = function(release)
+                    local version = release:gsub("^v", "")
+                    return ("google-java-format-%s.jar"):format(version)
```
williamboman, line 20: I think it'd be better to target the uber jar, otherwise users would have to ensure dependencies exist in the class path. Or is there some reason to use the slim jar?

```suggestion
                    return ("google-java-format-%s-all-deps.jar"):format(version)
```

williamboman (commented):
Cool! Just one question regarding the jar target

jay-babu:
> Cool! Just one question regarding the jar target

Yep, that makes sense

PR #961, opened 2023-02-01
feat(registry): add gospel

WhoIsSethDaniel:


WhoIsSethDaniel:
A question: this project has no packages. It also requires that `hunspell` be installed to successfully install (if `hunspell` is not installed `mason` just says "failed to install" -- if you use `go install` directly there is a more informative error message). I don't think there's a way to specify that a package requires some third-party dependency is there?

lua/mason-registry/gospel/init.lua
```diff
@@ -0,0 +1,11 @@
+local Pkg = require "mason-core.package"
+local go = require "mason-core.managers.go"
+
+return Pkg.new {
+    name = "gospel",
+    desc = [[misspelled word linter for Go comments, string literals and embedded files]],
+    homepage = "https://github.com/kortschak/gospel",
+    categories = {},
```
williamboman, line 8: This seems to be a linter, no?

```suggestion
    categories = { Pkg.Cat.Linter },
```

williamboman:
> A question: this project has no packages. It also requires that hunspell be installed to successfully install (if hunspell is not installed mason just says "failed to install" -- if you use go install directly there is a more informative error message). I don't think there's a way to specify that a package requires some third-party dependency is there?

There sort of is, but it's very imperative and basic. Hoping to standardize this soon in https://github.com/mason-org/mason-registry.

lua/mason-registry/gospel/init.lua
WhoIsSethDaniel: yes

WhoIsSethDaniel:
> There sort of is, but it's very imperative and basic (`grep std.ensure_executable`). Hoping to standardize this soon in https://github.com/mason-org/mason-registry.

so there's nothing to be done right now? Or there is? I'm mostly worried that people will try to install it, it will say it failed to install, and then will submit bug reports not realizing that the problem is on their end.

williamboman:
@williambotman /fixup

williamboman:
> so there's nothing to be done right now? Or there is? I'm mostly worried that people will try to install it, it will say it failed to install, and then will submit bug reports not realizing that the problem is on their end.

Let's skip it for now!

PR #971, opened 2023-02-06
feat(registry): add fennel-language-server

Massolari:


williamboman:
@williambotman /fixup

PR #1016, opened 2023-02-20
feat: configurable registries

williamboman:


A-Lamia:
I looked through and couldn't find anything but what is the code equivalent of this now ? `lrequire('mason-registry').get_package('codelldb'):get_install_path()`

williamboman:
It's the same, if that's erroring you're probably setting up mason too late.

A-Lamia:
yeah i thought that was the problem but now i don't think it is

A-Lamia:
Yeah you are right was not loading in the correct order, sorry for the annoyance :D

jcdickinson:
This is nuking 100% of my manual lsp configs, as I am using Mason only as a fallback when something isn't provided by nixpkgs/in the local env. Is that expected behavior?

PR #1040, opened 2023-02-28
feat(registry): ensure pyproject.toml support for pydocstyle

gabrielbdsantos:
Pydocstyle only supports `pyproject.toml` if the `toml` package is installed.


williamboman (approved):
:+1:

PR #1075, opened 2023-03-07
feat(registry): add bzl

zaucy:


zaucy:
I missed the mason-registry repository and since bzl isn't versioned I don't think it can be included. Sorry for the noise!

zaucy:
Thanks for the response! I'll look into the generic package type :)

williamboman:
Added in https://github.com/mason-org/mason-registry/pull/1016

PR #1078, opened 2023-03-08
php-cs-fixer: use new PHP-CS-Fixer/PHP-CS-Fixer repo

ueaner:
Also submitted a [PR#1437](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/1437) in the null-ls.nvim project


williamboman:
@williambotman /fixup

PR #1103, opened 2023-03-15
add gleam server

worming004:
add  gleam server reference


williamboman:
Hey! That file [shouldn't be edited manually](https://github.com/williamboman/mason.nvim/blob/2b7fb14fd46e659aab0ac1105ea6337e375e1be9/lua/mason/mappings/language.lua#L1-L2). There is also no Gleam package in Mason, the language server is bundled with `gleam` itself (see https://github.com/williamboman/mason.nvim/issues/884).

PR #1124, opened 2023-03-23
fix(scripts): setup mason and init registries

williamboman:


williamboman:
@williambotman /fixup

PR #1127, opened 2023-03-23
chore: migrate lua-language-server

williamboman:


mawkler:
@williamboman After updating Mason today I started getting the message `[mason-lspconfig.nvim] Server "lua_ls" is not a valid entry in ensure_installed. Make sure to only provide lspconfig server names.` on startup. Is this expected?

PR #1135, opened 2023-03-25
refactor: download checksum after unzip

williamboman:


williamboman:
@williambotman /fixup

PR #1146, opened 2023-03-28
feat(registry): add ability to register package aliases

avegancafe:
This is an attempt at resolving this issue https://github.com/williamboman/mason-lspconfig.nvim/issues/116. Let me know if there's a better/preferred way to do this!

### DIFF
--- a/lua/mason-registry/init.lua
+++ b/lua/mason-registry/init.lua
@@ -209,7 +209,7 @@ end
 
 ---@param name string
 function M.get_package_aliases(name)
-    return aliases[name]
+    return aliases[name] or {}
 end
 
 return M
--- a/lua/mason-registry/init.lua
+++ b/lua/mason-registry/init.lua
@@ -193,7 +193,7 @@ function M.refresh(cb)
     end
 end
 
----table<string, string[]>
+---@type table<string, string[]>
 local aliases = {}
 
 ---Register aliases for the specified packages
--- a/lua/mason-registry/init.lua
+++ b/lua/mason-registry/init.lua
@@ -193,4 +193,23 @@ function M.refresh(cb)
     end
 end
 
+---table<string, string[]>
+local aliases = {}
+
+---Register aliases for the specified packages
+---@param new_aliases table<string, string[]>
+function M.register_aliases(new_aliases)
+    for key, value in pairs(new_aliases) do
+        aliases[key] = aliases[key] or {}
+        for _, alias in pairs(value) do
+            table.insert(aliases[key], alias)
+        end
+    end
+end
+
+---@param name string
+function M.get_aliases(name)
+    return aliases[name]
+end
+
 return M
--- a/lua/mason-registry/init.lua
+++ b/lua/mason-registry/init.lua
@@ -199,10 +199,12 @@ local aliases = {}
 ---Register aliases for the specified packages
 ---@param new_aliases table<string, string[]>
 function M.register_package_aliases(new_aliases)
-    for key, value in pairs(new_aliases) do
-        aliases[key] = aliases[key] or {}
-        for _, alias in pairs(value) do
-            table.insert(aliases[key], alias)
+    for pkg_name, pkg_aliases in pairs(new_aliases) do
+        aliases[pkg_name] = aliases[pkg_name] or {}
+        for _, alias in pairs(pkg_aliases) do
+            if alias ~= pkg_name then
+                table.insert(aliases[pkg_name], alias)
+            end
         end
     end
 end
--- a/lua/mason-registry/init.lua
+++ b/lua/mason-registry/init.lua
@@ -198,7 +198,7 @@ local aliases = {}
 
 ---Register aliases for the specified packages
 ---@param new_aliases table<string, string[]>
-function M.register_aliases(new_aliases)
+function M.register_package_aliases(new_aliases)
     for key, value in pairs(new_aliases) do
         aliases[key] = aliases[key] or {}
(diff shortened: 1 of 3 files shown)


williamboman (commented):
Hey! I appreciate the initiative!! One thing that I'd like to do is keep things decoupled, and keep mason.nvim blissfully unaware of external concepts. I'd like to invert the responsibility, and instead provide facilities in mason.nvim that allow 3rd parties to provide additional metadata about packages.

What I had in mind was introducing the concept of package aliases, which could be registered like so:

```lua
require("mason-registry").register_aliases {
  ["lua-language-server"] = { "lua_ls" }
}
```

and then in `mason-lspconfig.nvim`'s `setup()` function, these aliases would be registered (the mapping already exist):

```lua
function M.setup()
  -- …
  require("mason-registry").register_aliases(...)
end
```

avegancafe:
Ohh I see I'll give that a stab!

lua/mason-registry/init.lua
```diff
@@ -193,4 +193,23 @@ function M.refresh(cb)
     end
 end
 
+---table<string, string[]>
+local aliases = {}
```
avegancafe, line 197: Happy to implement this however you'd like, I figured it might be beneficial from an API perspective for the actual mapping to be private

williamboman: Looks good to me!

lua/mason-registry/init.lua
```diff
@@ -193,4 +193,23 @@ function M.refresh(cb)
...
+
+---Register aliases for the specified packages
+---@param new_aliases table<string, string[]>
+function M.register_aliases(new_aliases)
```
williamboman, line 201: Let's rename these functions to narrow it down to packages specifically (future proofing)

```suggestion
function M.register_package_aliases(new_aliases)
```

avegancafe: done! I also renamed `get_aliases` to `get_package_aliases` to match

lua/mason-registry/init.lua
```diff
@@ -193,4 +193,23 @@ function M.refresh(cb)
...
+        aliases[key] = aliases[key] or {}
+        for _, alias in pairs(value) do
+            table.insert(aliases[key], alias)
+        end
+    end
+end
+
+---@param name string
+function M.get_package_aliases(name)
+    return aliases[name]
```
williamboman, line 212: I think it'd be nice for this function to always return a table:

```suggestion
    return aliases[name] or {}
```

lua/mason-registry/init.lua
```diff
@@ -193,4 +193,23 @@ function M.refresh(cb)
     end
 end
 
+---table<string, string[]>
```
williamboman, line 196:
```suggestion
---@type table<string, string[]>
```

lua/mason-registry/init.lua
```diff
@@ -193,4 +193,23 @@ function M.refresh(cb)
...
+local aliases = {}
+
+---Register aliases for the specified packages
+---@param new_aliases table<string, string[]>
+function M.register_aliases(new_aliases)
+    for key, value in pairs(new_aliases) do
+        aliases[key] = aliases[key] or {}
+        for _, alias in pairs(value) do
```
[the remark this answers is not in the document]
williamboman: The non-imperative way would perhaps be to add a new `_.merge_with` function to the functional modules, which would allow for something like:

```lua
function M.register_package_aliases(new_aliases)
    aliases = _.merge_with(_.concat, aliases, new_aliases)
end
```

I think this looks good to me though!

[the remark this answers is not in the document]
avegancafe: Oh fantastic, definitely prefer to do this declaratively

williamboman:
What do you think about inlining the alias besides the name, making it always visible? It's what I had in mind at least, I feel like it belongs there 🤷. I've noticed a lot of people search the buffer like they'd normally do with `/`, so having aliases always visible would also aid with searching for packages

Screenshot 2023-04-05 at 21 10 26

lua/mason-registry/init.lua
```diff
@@ -193,4 +193,23 @@ function M.refresh(cb)
...
+---table<string, string[]>
+local aliases = {}
+
+---Register aliases for the specified packages
+---@param new_aliases table<string, string[]>
+function M.register_package_aliases(new_aliases)
+    for key, value in pairs(new_aliases) do
+        aliases[key] = aliases[key] or {}
+        for _, alias in pairs(value) do
+            table.insert(aliases[key], alias)
```
williamboman, line 205: I noticed another thing: I don't think we should apply the alias in case it correspond with the package name already as it's not really an alias that point (it also makes the UI look a bit weird with duplicate names).

```suggestion
    for pkg_name, pkg_aliases in pairs(new_aliases) do
        aliases[pkg_name] = aliases[pkg_name] or {}
        for _, alias in pairs(value) do
            if alias ~= pkg_name then
              table.insert(aliases[pkg_name], alias)
            end
```

avegancafe: Nice yeah that makes sense to me. Done

lua/mason-registry/init.lua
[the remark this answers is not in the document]
avegancafe: With the new feature of omitting aliases from the list if they match the mason package name it's a little wonky to do it declaratively now actually. Going to resolve this for now, happy to take another stab at it though if you think it's better

williamboman (approved):
Awesome, thanks 🚀!!

PR #1185, opened 2023-04-06
fix: healthcheck completion of mason

dundargoc:
When tab-completing :checkhealth it will show "mason.health", which will result in the error
"ERROR No healthcheck found for "mason/health" plugin".


dundargoc:
This is a limitation that should arguably be fixed in neovim core, but meanwhile this should so IMO.

williamboman (approved):
Ah thanks, I always wondered why it did that

PR #1213, opened 2023-04-14
ci: introducing semantic versioning to mason.nvim

jay-babu:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,103 @@
...
+        - cbfmt
+        - selene
+        - stylua
+        - tests
+      runs-on: ubuntu-latest
+      steps:
+        - uses: google-github-actions/release-please-action@v3
+          id: release
+          with:
+            token: ${{ secrets.RELEASE_TOKEN }}
```
williamboman, line 94: Can we use the default secret (`GITHUB_TOKEN`) instead (i.e. omit this entirely)?

williamboman (commented):
Sweet! Any reason to merge all workflows into one? I think it'd be neat to keep them separated. I think it's fine if the release job doesn't explicitly depend on the other jobs, as the `main` branch is protected anyway.

It also seems like the release job needs some extra permissions, so it'd be they were only applied for that job (it's probably possible to configure this on a per-job basis though):

```yaml
permissions:
  contents: write
  pull-requests: write
```

.github/workflows/ci.yml
jay-babu: yes, sorry. bad copy paste

jay-babu:
> Sweet! Any reason to merge all workflows into one? I think it'd be neat to keep them separated. I think it's fine if the release job doesn't explicitly depend on the other jobs, as the `main` branch is protected anyway.
> 
> It also seems like the release job needs some extra permissions, so it'd be they were only applied for that job (it's probably possible to configure this on a per-job basis though):
> 
> ```yaml
> permissions:
>   contents: write
>   pull-requests: write
> ```

okay, if you are fine with not having the release job depend on the others, then they can be separate

williamboman:
Sweet, let's give this a try

Issue #1284, opened 2023-05-03
Allow passing in registry information in setup instead of reaching out to mason-registry.dev


reegnz:
### I've searched open issues for similar requests

- [x] Yes

### Is your feature request related to a problem? Please describe.

I just noticed that information on how to download packages got removed from the repo and moved to a SaaS URL in #1246.

Arguments against the approach were already made in https://github.com/williamboman/mason.nvim/issues/582, and that change therefore is very confusing now.

### Describe the solution you'd like

Please provide an option to download that information from any alternate URL, or even allow mason to consume the list in it's `setup`, so one could still continue using a lua module to consume the list.

reegnz:
I'm seeing contradicting information in the repo:
https://github.com/williamboman/mason.nvim/blob/main/PACKAGES.md says it got moved to mason-registry.dev, but then the docs say it's fetched from `github:mason-org/mason-registry`: https://github.com/williamboman/mason.nvim#default-configuration

So which one is it?

I'm guessing mason fetches from [github](https://github.com/mason-org/mason-registry) so I can fix the PACKAGES.md to point to the github repo and not the domain, if it actually fetches package info from github.

williamboman:
PACKAGES.md used to contain a list of the core packages. This was moved to https://mason-registry.dev/registry/list (https://github.com/mason-org/public-web) after the package definitions were migrated to https://github.com/mason-org/mason-registry/. The PACKAGES.md file was kept because it's one of the most visited files according to GitHub's repo traffic data, so what remains is essentially just a "301 redirect". When/if traffic to that file decreases I'll remove it altogether as it no longer serves a purpose.

Contribution guidelines are available in https://github.com/williamboman/mason.nvim/blob/main/CONTRIBUTING.md#adding-a-new-package and some brief information about registries [are in the README](https://github.com/williamboman/mason.nvim#registries) as well.

PR #1285, opened 2023-05-03
Update PACKAGES.md to point to default registry repo

reegnz:
When looking at the markdown it was very confusing where the packages are coming from now. This allows pointing future contributors to the right place.

Fixes confusion in #1284


williamboman:
The purpose of this file was/is to provide a list of available packages

Issue #1415, opened 2023-07-17
tar check fails on openbsd


js0000:
### I've searched open issues for similar requests

- [X] Yes

### Is your feature request related to a problem? Please describe.

mason.nvim is great- thx!

in checkhealth i get this:

- ERROR tar: not available
  - ADVICE:
    - spawn: tar failed with exit code 1 and signal 0. tar: unknown option -- -
      usage: tar {crtux}[014578befHhjLmNOoPpqsvwXZz]
      [blocking-factor | archive | replstr] [-C directory] [-I file]
      [file ...]
      tar {-crtux} [-014578eHhjLmNOoPpqvwXZz] [-b blocking-factor]
      [-C directory] [-f archive] [-I file] [-s replstr] [file ...]

i am guessing this is because openbsd's tar executable does not use the same arguments as the gnu version of tar. i can install the gnu version (called "gtar"), but i am not at all sure if there is any way for mason to know this (maybe search for gtar if tar is not found ... ?)

### Describe the solution you'd like

for mason to check for gtar if tar does not support the necessary options

### Describe potential alternatives you've considered

i've thought about digging through the mason source, but if i don't have to, i'd rather not ...

### Additional context

i am running openbsd 7.3

williamboman:
Hello! Is this only failing the health check or is it failing to unpack files as well (lua-language-server for example)?

3uryd1ce:
> Hello! Is this only failing the health check or is it failing to unpack files as well (lua-language-server for example)?

Hi @williamboman, I also use OpenBSD (I'm on 7.3-current) and can confirm that `gtar` isn't found. I'm also unable to install `lua-language-server` (along with some others: `shfmt`, `shellcheck`, ...). `lua-language-server` installation fails with this output:

```
[INFO  Fri Jul 28 19:35:43 2023] ...e/nvim/lazy/mason.nvim/lua/mason-core/installer/init.lua:184: Executing installer for Package(name=lua-language-server) {}
[DEBUG Fri Jul 28 19:35:43 2023] ...e/nvim/lazy/mason.nvim/lua/mason-core/installer/init.lua:58: Attempting to lock package Package(name=lua-language-server)
[DEBUG Fri Jul 28 19:35:43 2023] ...e/nvim/lazy/mason.nvim/lua/mason-core/installer/init.lua:70: Wrote lockfile Package(name=lua-language-server)
[...]

PR #1433, opened 2023-07-31
fix(std): use gtar if available

3uryd1ce:
Addresses #1415.

These changes are needed because the relevant code currently assumes GNU tar features like automatic decompression and the `--no-same-owner` flag, but doesn't detect and use GNU tar on platforms where it's named gtar. This causes a number of failures.

With these changes, the health check passes and tarballs are correctly unpacked on platforms other than Linux. This fix affects at least OpenBSD, and probably other systems like FreeBSD, NetBSD, DragonFly BSD, etc.

tests(std_spec.lua): add a local function named get_tar_cmd() that returns the correct version of tar to call. Adjust the tests so that they call gtar when appropriate.

docs(README.md): update requirements so that they explicitly state that GNU tar is used. Clean up requirements by placing them in separate bullet points, rather than listing them in one line.

### DIFF
--- a/lua/mason-core/installer/managers/std.lua
+++ b/lua/mason-core/installer/managers/std.lua
@@ -109,6 +109,20 @@ end
 local function untar(rel_path)
     log.fmt_debug("std: untar %s", rel_path)
     local ctx = installer.context()
+    local tar = vim.fn.executable "gtar" == 1 and "gtar" or "tar"
+    return ctx.spawn[tar]({ "--no-same-owner", "-xvf", rel_path }):on_success(function()
+        pcall(function()
+            ctx.fs:unlink(rel_path)
+        end)
+    end)
+end
+
+---@async
+---@param rel_path string
+---@nodiscard
+function M.untar(rel_path)
+    log.fmt_debug("std: untar %s", rel_path)
+    local ctx = installer.context()
     a.scheduler()
     local tar = vim.fn.executable "gtar" == 1 and "gtar" or "tar"
     return ctx.spawn[tar]({ "--no-same-owner", "-xvf", rel_path }):on_success(function()
--- a/lua/mason-core/installer/managers/std.lua
+++ b/lua/mason-core/installer/managers/std.lua
@@ -109,20 +109,6 @@ end
 local function untar(rel_path)
     log.fmt_debug("std: untar %s", rel_path)
     local ctx = installer.context()
-    local tar = vim.fn.executable "gtar" == 1 and "gtar" or "tar"
-    return ctx.spawn[tar]({ "--no-same-owner", "-xvf", rel_path }):on_success(function()
-        pcall(function()
-            ctx.fs:unlink(rel_path)
-        end)
-    end)
-end
-
----@async
----@param rel_path string
----@nodiscard
-function M.untar(rel_path)
-    log.fmt_debug("std: untar %s", rel_path)
-    local ctx = installer.context()
     a.scheduler()
     local tar = vim.fn.executable "gtar" == 1 and "gtar" or "tar"
     return ctx.spawn[tar]({ "--no-same-owner", "-xvf", rel_path }):on_success(function()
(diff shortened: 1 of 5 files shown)


lua/mason-core/installer/managers/std.lua
```diff
@@ -108,7 +108,11 @@ end
 local function untar(rel_path)
     log.fmt_debug("std: untar %s", rel_path)
     local ctx = installer.context()
-    return ctx.spawn.tar({ "--no-same-owner", "-xvf", rel_path }):on_success(function()
+    local tar_cmd = ctx.spawn.tar
+    if platform.is.unix and not platform.is.linux then
+        tar_cmd = ctx.spawn.gtar
+    end
```
3uryd1ce, line 114: Sections like this with the platform check would be nice to refactor into a function somewhere else (but I don't know how you would want to organize this so I've left them as is for now). I see this as being useful because if the code needed to change for some reason, it could be changed in one place.

lua/mason-core/managers/std/init.lua
```diff
@@ -100,7 +100,11 @@ end
 function M.untar(file, opts)
     opts = opts or {}
     local ctx = installer.context()
-    ctx.spawn.tar {
+    local tar_cmd = ctx.spawn.tar
+    if platform.is.unix and not platform.is.linux then
+        tar_cmd = ctx.spawn.gtar
+    end
+    tar_cmd {
```
williamboman, line 107: This module and all others inside `mason-core.managers` are legacy and not used anywhere in mason.nvim. I think we can leave these as-is seeing as they're no longer supported anyways.

lua/mason-core/installer/managers/std.lua
```diff
@@ -108,7 +108,11 @@ end
...
+    return tar_cmd({ "--no-same-owner", "-xvf", rel_path }):on_success(function()
```
williamboman, line 115: I think we can simplify this to simply prefer `gtar` if it exists and if not fall back to `tar`. That way we can avoid relying on broader platform detection and all the quirks it may incur. Something like the following ought to do it.

```lua
local a = require("mason-core.async")

-- …

function M.untar()
  -- …
  a.scheduler()
  local tar = vim.fn.executable("gtar") == 1 and "gtar" or "tar"
  return ctx.spawn[tar] { ... }
end
```

The `a.scheduler()` (also notice the additional import) is required because this function may execute outside Neovim's main loop, in which case `vim.fn` functions cannot be called.

tests/mason-core/installer/managers/std_spec.lua
```diff
@@ -1,8 +1,16 @@
...
 local std = require "mason-core.installer.managers.std"
 local stub = require "luassert.stub"
 
+local function get_tar_cmd(ctx)
+    if platform.is.unix and not platform.is.linux then
+        return ctx.spawn.gtar
+    end
+    return ctx.spawn.tar
+end
+
```
williamboman, line 13: By leveraging `vim.fn.executable("gtar")` we can keep these tests the same they were before without having to incorporate any logic.

```lua
describe("...", function ()
  before_each(function ()
    stub(vim.fn, "executable")
    vim.fn.executable.on_call_with("gtar").returns(0)
  end)
end)
```

Then we can also add a dedicated test for gtar detection:

```lua
it("should use gtar if available", function ()
  -- … boilerplate
  stub(vim.fn, "executable")
  vim.fn.executable.on_call_with("gtar").returns(1)
  -- … execute & assert
end)
```

williamboman (commented):
Thanks for the PR! Some comments.

3uryd1ce:
Hi @williamboman, I made some more commits. I don't know that much Lua, so I was unsure how to do certain things. I did my best, I hope my updates are close to what you had in mind.

csarath:
Is there a way to use tar by default. I'm facing "gtar: This does not look like a tar archive" with many of lsp servers.

PR #1557, opened 2023-11-15
feat: [pip] added venv_args to configuration options

magnuslarsen:
Added the ability to pass arguments to the `venv`-creator

This is especially useful when for example, using `python-language-server` on a Linux based distro that packages pip-packages as system packages (Debian-based, Arch, ...)

By default you do not get access to system-site packages, meaning no LSP-features for those packages, unless you manually install them (again) in the `python-language-server`-venv

Passing the following circumvents all of that, by allowing the venv to access system packages:
```lua
require("mason").setup({
	pip = {
		venv_args = { "--system-site-packages" },
	},
})
```

ref: https://docs.python.org/3/library/venv.html#creating-virtual-environments

### DIFF
--- a/lua/mason-core/installer/managers/pypi.lua
+++ b/lua/mason-core/installer/managers/pypi.lua
@@ -6,6 +6,7 @@ local installer = require "mason-core.installer"
 local log = require "mason-core.log"
 local path = require "mason-core.path"
 local platform = require "mason-core.platform"
+local settings = require "mason.settings"
 
 local M = {}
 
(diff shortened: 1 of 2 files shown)


williamboman (commented):
Hey, thanks for the PR :+1:!

This has been brought up before and I'm a bit hesitant to the idea. My rationale is that, per my understanding, this would only really play nicely if you install your packages globally (according to whatever `--system-site-packages` considers global). If you choose to use a project-local virtual environment (which I'd like to believe is the norm), local packages would not be recognized even with this flag enabled. At this point you might as well manage this package outside of Mason, along with the rest.

Furthermore, python-lsp-server in particular should support loading different environments [via its configuration](https://github.com/williamboman/mason.nvim/issues/1276#issuecomment-1533884170), so I'd like to believe this is mostly a configuration issue. I'm aware this is severely undocumented and probably causes issues for many users, I'm not sure where the best place to solve it would be though.

magnuslarsen:
You are right, this seems to be a configuration issue.

I'm using the community version of the python-language-server (pylsp), but the same config as mentioned in above issue worked here as well

I got a few dependency errors after setting the environment to the /usr/bin/python3 (likely due to Ubuntu shipping old pip packages) - nothing that prevents it from working though

Thanks for an invaluable plugin :-)

BlueDrink9:
Responded to this argument in #1584

PR #1564, opened 2023-11-25
fix(registry): fix yq usage

D0ot:
I get this error message after `:Mason`. It seems that the yq usage is not right. 

```
  Uninstalled registries
  Packages from the following registries are unavailable. Press C to install.
   - local: /home/doot/tmp/mason-registry [uninstalled]
  
  Registry installation failed with the following error:
    FileRegistrySource(path=/home/doot/tmp/mason-registry) failed to install: ...nvim/lazy/mason.nvim/lua/mason-registry/sources/file.lua:128: ...nvim/lazy/mason.nvim/lua/mason-registry/sources/file.lua:120: Failed to parse /home/doot/tmp/mason-registry/packages/erlang-ls/package.yaml.
 ```

After this fix, it works as expected.

### DIFF
--- a/lua/mason-registry/sources/file.lua
+++ b/lua/mason-registry/sources/file.lua
@@ -108,8 +108,7 @@ function FileRegistrySource:install()
                     local yaml_spec = fs.async.read_file(package_file)
                     local spec = vim.json.decode(spawn
                         [yq]({
-                            "-o",
-                            "json",
+                            ".",
                             on_spawn = a.scope(function(_, stdio)
                                 local stdin = stdio[1]
                                 async_uv.write(stdin, yaml_spec)

### CODE REVIEW

**Comment by D0ot** [comment]
My mason config:

```lua
require("mason").setup({
	ui = {
		icons = {
			package_installed = "✓",
			package_pending = "➜",
			package_uninstalled = "✗",
		},
	},
	registries = {
		"file:/home/doot/tmp/mason-registry",
	},
})
```

williamboman:
What does `yq .` do and how does it differ from `yq -o json`? There might be more information in `:MasonLog` in terms of what is going wrong, however I'm not sure to what extent errors that happen here are logged

D0ot:
Sorry, I check my yq version, it is too old. and it does not support `-o json` options...

williamboman:
Ok I see! Just curious, where and how did you find out about the `file:` registry protocol? I've not documented it yet primarily because of the dependency on `yq` (which you can install via Mason through the default registry settings).

D0ot:
> Ok I see! Just curious, where and how did you find out about the `file:` registry protocol? I've not documented it yet primarily because of the dependency on `yq` (which you can install via Mason through the default registry settings).

The thing was that I wanted to use mason.nvim on a server which does not have access to internet and I decided to modify mason.nvim for my no-internet needs. Then I read some code about `registry` and `provider`and the `file:` registry protocol was found and I just tried it.

PR #1581, opened 2023-12-19
feat(npm): add `install_args` option

amittamari:
(Rework of #1580)

Often, when I'm working, my npm registry is set to the work's registry, which sometimes causes npm installations in Mason to fail.

To overcome this, I added the option to provide `install_args` to npm, the same as available in pip, so I can configure Mason to always use the default npm registry.

Note: I wasn't sure if I should also support it in `lua/mason-core/managers/npm/init.lua`, because I could not find any references where it is used (except tests). Let me know, and I can add it if needed.


amittamari:
Resolved conflicts

devlopersabbir (approved):

amittamari:
@williamboman can you please give this one a review? 🙏🏻 It's a relatively small change

williamboman (approved):
Hey, terribly sorry for so late review. I saw this when you first created it but I wanted to avoid exposing this as a setting and find a different `.npmrc` based approach (I never liked the equivalent `pip.install_args` setting either as it exposes such tight coupling to implementation details). I think in the absence of another solution this makes sense

PR #1585, opened 2023-12-29
fix(cargo): don't attempt to fetch versions when version targets commit SHA

williamboman:


williamboman:
@williambotman /fixup

PR #1597, opened 2024-01-11
feat(ui): remove whitespace in empty lines

### LINKED ISSUE #1529: feat(ui): Delete Whitespaces in Blankline

- [X] Yes

First of all, thanks for the plugin. I am having amazing experience so far but got a very minor hiccup.

I would like to quickly move the cursor to the `Available` section.
I believe the fastest way would be to use the `}` keybind to move down a paragraph, but since there are whitespaces between the `Installed` section, I cannot move between the sections easily.

As you can see, there are whitespaces (highlighted in red) above the `Available` section which does not make it a separate paragraph.
image

It would be nice if there were no whitespaces in the blankline between the sections.

Or, another solution would be to add a specific keybind (e.g. `[[`, `]]`) to jump to the next section.

I believe changing some code near this part (or inside `get_styles`) so that `indentation` is not added if `line:len() == 0` or something similar to that would be a possible fix.
But I was hesitant to submit a PR becuz this method might rather be a _hack_ and there might be a better and more elegant way to solve the issue.

https://github.com/williamboman/mason.nvim/blob/cd7835b15f5a4204fc37e0aa739347472121a54c/lua/mason-core/ui/display.lua#L122-L133

### Additional context

_No response_

WillLillis:
This PR prevents indentation from being added to blank lines in the UI, which in turn allows the user to move between sections using the `}` keybind. Here's an example gif:

mason_whitespace

Closes #1529

### DIFF
--- a/lua/mason-core/ui/display.lua
+++ b/lua/mason-core/ui/display.lua
@@ -119,10 +119,9 @@ local function render_node(viewport_context, node, _render_context, _output)
                 end
             end
 
-            local active_styles = get_styles(full_line, render_context)
-
-            -- apply indentation to non-empty lines
+            -- only apply cascading styles to non-empty lines
             if string.len(full_line) > 0 then
+                local active_styles = get_styles(full_line, render_context)
                 full_line = (" "):rep(active_styles.indentation) .. full_line
                 for j = 1, #line_highlights do
                     local highlight = line_highlights[j]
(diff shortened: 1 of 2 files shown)


williamboman (approved):
Makes sense, thanks! I added a test case and made a slight adjustment so `get_styles` is only called when it's actually needed

PR #1598, opened 2024-01-11
docs: add information that `vim.g.python3_host_prog` is used

chrisgrieser:
I recently had some headaches wondering mason was using an outdated python version. The recent version of some packages available via mason require a more recent python, but as mason was using an older python version, they would not be able to be installed.

I figured out that it was using my system's python, and with some research figured out that it is in fact `vim.g.python3_host_prog` that determines which python is used by mason. I think it is worth mentioning in the README that this variable determines what python mason uses. 

(Maybe it could also make sense to have a setting in the mason config to explicitly tell mason which python it should use?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,10 @@ local DEFAULT_SETTINGS = {
 }
 ```
 
+> [!NOTE]
+> mason uses the python version specified via `vim.g.python3_host_prog`. This can be relevant for some packages,
+> where the most recent version requires a more recent python version than the one installed system's default.
+
 ---
 
 <sup>


williamboman (commented):
Hey, thanks! I'm actually not sure why it's checking that variable for python. I think I figured it'd improve the chances of finding a python executable should a user not have it in `PATH`. It's however inconsistent with how other system dependencies are resolved - I think we should remove usage of `vim.g.python3_host_prog` entirely. I'll open a separate PR for this!

williamboman:
Superseded by https://github.com/williamboman/mason.nvim/pull/1606

Issue #1617, opened 2024-02-02
Can support a custom callback function for detecting server updates?


fcying:
### I've searched open issues for similar requests

- [X] Yes

### Is your feature request related to a problem? Please describe.

no

### Describe the solution you'd like

I register a custom server, the server update from local intranet, 
It can be used fine, but it cannot detect update.
Is it possible to add a 'custom' type in the `get_new_version_by_type` array? When using this type, it will invoke a custom callback function to detect update.

PR #1623, opened 2024-02-06
feat: add type `custom` for custom server check version

fcying:
for #1617


williamboman (commented):
Hello and thanks for the PR! These parts are unused internally since 1.0.0. They still remain in order to not break any potential 3rd party usage but will soon be removed entirely (I also can't remember opening up these APIs for 3rd party use). I'd rather not add more to these parts.

PR #1643, opened 2024-03-02
feat(linker): use relative targets for exec scripts

### LINKED ISSUE #1156: Support linking executables using relative symlinks

Yes

Idealy I would like to be able to just copy a nvim setup with mason to a different machine and it should just work.

Currently the symlinks mason creates for executables in bin/ are not relative. This makes it hard to make a portable neovim setup as most of the times the path will be different on other machines.

```
[rs@arch bin]$ ls -l
total 16
lrwxrwxrwx 1 rs rs 47 mar 30 16:23 jq -> /home/rs/.local/share/nvim/mason/packages/jq/jq
lrwxrwxrwx 1 rs rs 81 mar 30 15:24 lua-language-server -> /home/rs/.local/share/nvim/mason/packages/lua-language-server/lua-language-server
lrwxrwxrwx 1 rs rs 77 mar 30 16:20 prettier -> /home/rs/.local/share/nvim/mason/packages/prettier/node_modules/.bin/prettier
``` 

 

The symlinks should be relative to the bin/ directory.

```
[rs@arch bin]$ ls -l
total 16
lrwxrwxrwx 1 rs rs 47 mar 30 16:23 jq -> ../packages/jq/jq
lrwxrwxrwx 1 rs rs 81 mar 30 15:24 lua-language-server -> ../packages/lua-language-server/lua-language-server
lrwxrwxrwx 1 rs rs 77 mar 30 16:20 prettier -> ../packages/prettier/node_modules/.bin/prettier
```

_No response_

### Additional context

_No response_

henry-hsieh:
Closes #1156 and improves #1525.

### DIFF
--- a/lua/mason-core/installer/context/init.lua
+++ b/lua/mason-core/installer/context/init.lua
@@ -168,10 +168,7 @@ function InstallContext:write_exec_wrapper(new_executable_rel_path, target_execu
     end
     return self:write_shell_exec_wrapper(
         new_executable_rel_path,
-        ("%q"):format(path.concat {
-            self.package:get_install_path(),
-            target_executable_rel_path,
-        })
+        target_executable_rel_path
     )
 end
 
@@ -183,6 +180,7 @@ exec %s "$@"
 
 local BATCH_TEMPLATE = _.dedent [[
 @ECHO off
+SET scriptpath=%~dp0
 %s
 %s %%*
 ]]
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Thanks for the PR! I've however already implemented a similar fix locally! It's largely the same except that the resulting wrapper executable can be executed from any context - I believe this solution would make the wrapper script error if executed directly (it has to be executed from the symlink in `bin/` for the path resolution to function).

henry-hsieh:
> Thanks for the PR! I've however already implemented a similar fix locally! It's largely the same except that the resulting wrapper executable can be executed from any context - I believe this solution would make the wrapper script error if executed directly (it has to be executed from the symlink in `bin/` for the path resolution to function).

Yes, that's true. Hope the fix can be uploaded soon!

Issue #1646, opened 2024-03-05
Checkhealth fails to report luarocks version number


FerretDetective:
### I've searched open issues for similar requests

- [X] Yes

### I've recently downloaded the latest plugin version of mason.nvim

- [X] Yes

### Problem description

When using a luarocks version with a minor version number which has more than one digit mason will fail to parse the version number.

### Expected behavior

Mason should parse the version number of luarocks regardless of how many digits the minor version number is.

### Steps to reproduce

1. Install luarocks with a minor version number that has more than one digit (exg: 3.10.0)
2. run `:checkhealth mason`

### Neovim version (>= 0.7)

NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1702233742

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info

### Operating system/version

Linux archlinux 6.7.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 03 Mar 2024 00:30:36 +0000 x86_64 GNU/Linux

### Healthcheck output

```shell
mason: require("mason.health").check()

mason.nvim ~
- OK mason.nvim version v1.10.0
- OK PATH: prepend
- OK Providers: 
  mason.providers.registry-api
  mason.providers.client
- OK neovim version >= 0.7.0

mason.nvim [Registries] ~
- OK Registry `github.com/mason-org/mason-registry version: 2024-03-04-scarce-bonito` is installed.

mason.nvim [Core utils] ~
- OK unzip: `UnZip 6.00 of 20 April 2009, by Info-ZIP.  Maintained by C. Spieler.  Send`
- OK wget: `GNU Wget 1.21.4 built on linux-gnu.`
- OK curl: `curl 8.6.0 (x86_64-pc-linux-gnu) libcurl/8.6.0 OpenSSL/3.2.1 zlib/1.3.1 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.7 libpsl/0.21.2 libssh2/1.11.0 nghttp2/1.60.0 nghttp3/1.2.0`
- OK gzip: `gzip 1.13`
- OK tar: `tar (GNU tar) 1.35`
- OK bash: `GNU bash, version 5.2.26(1)-release (x86_64-pc-linux-gnu)`
- OK sh: `Ok`

mason.nvim [Languages] ~
- OK Go: `go version go1.22.0 linux/amd64`
[...]

PR #1648, opened 2024-03-07
fix(health): support multidigit luarocks version numbers (#1646)

FerretDetective:
As described in #1646, this would fix the current inability for `checkhealth mason` to report luarocks versions with multidigit minor version numbers.


anjos:
With the new version of luarocks (3.10 and 3.11), it would be great to have this patch integrated on the main branch when possible!

williamboman (approved):
Oops, thanks!

PR #1650, opened 2024-03-14
fix: opam update before lsp upgrade

mercxry:
Hello 👋

I'm fixing a bug where installing a new version of `ocaml-lsp` would make `opam` complain about not being able to find the new version locally, this is because `opam update` needs to be run first to update the local repositories, so I'm running that command before any installation.

### DIFF
--- a/lua/mason-core/installer/managers/opam.lua
+++ b/lua/mason-core/installer/managers/opam.lua
@@ -15,13 +15,20 @@ function M.install(package, version)
     log.fmt_debug("opam: install %s %s", package, version)
     local ctx = installer.context()
     ctx.stdio_sink.stdout(("Installing opam package %s@%s…\n"):format(package, version))
-    return ctx.spawn.opam {
-        "install",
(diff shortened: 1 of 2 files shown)


williamboman (commented):
Thanks ⭐!

williamboman:
@williambotman /fixup

williamboman:
Ugh stylua is complaining, I'll fix it shortly.

I just saw the following in their docs which I don't understand, but which might be a bit problematic if it can't be turned off. Will have to research it a bit more

> It will also update any packages that are bound to version-controlled sources.

williamboman:
Hi again, I've not been able to find a resolution to the concern above of this command potentially modifying opam packages installed elsewhere on the system. One of the core principles of mason.nvim is to never touch the file system outside of Mason's root directory (except logs, cache, etc. that dependencies write). I'll go ahead and close this for now but I'd be open if we can resolve it somehow.

PR #1690, opened 2024-04-25
fix: replace deprecated table methods for neovim v0.10

skoch13:
Credits go to https://github.com/sindrets/diffview.nvim/pull/489

Fixes https://github.com/williamboman/mason.nvim/issues/1689

### DIFF
--- a/lua/mason-core/installer/registry/util.lua
+++ b/lua/mason-core/installer/registry/util.lua
@@ -15,7 +15,7 @@ end
 
 function M.flatten()
     return vim.fn.has "nvim-0.10" == 1 and function(...)
-        return vim.iter(...):flatten():totable()
+        return vim.iter(...):flatten(math.huge):totable()
     end or vim.tbl_flatten ---@diagnostic disable-line: deprecated
 end
 
--- a/lua/mason-core/installer/registry/util.lua
+++ b/lua/mason-core/installer/registry/util.lua
@@ -7,12 +7,24 @@ local platform = require "mason-core.platform"
 
 local M = {}
 
+---Helper functions for deprecated methods in nightly
+---@return function
+function M.islist()
+    return vim.fn.has "nvim-0.10" == 1 and vim.islist or vim.tbl_islist ---@diagnostic disable-line: deprecated
+end
+
+function M.flatten()
+    return vim.fn.has "nvim-0.10" == 1 and function(...)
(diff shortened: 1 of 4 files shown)


mikesmithgh:
FYI @skoch13 https://github.com/sindrets/diffview.nvim/pull/489#issuecomment-2085206944

I also do not see the error on the latest version of nvim nightly. 

I made a change to https://github.com/sindrets/diffview.nvim/pull/489 to add the depth of `math.huge` to the flatten function that should be updated in the PR.

skoch13:
@mikesmithgh thank you, I did mention that in the issue here https://github.com/williamboman/mason.nvim/issues/1689#issuecomment-2078711997

I'll update the PR accordingly, thank you

lua/mason-core/installer/registry/util.lua
```diff
@@ -7,12 +7,24 @@ local platform = require "mason-core.platform"
 
 local M = {}
 
+---Helper functions for deprecated methods in nightly
+---@return function
+function M.islist()
+    return vim.fn.has "nvim-0.10" == 1 and vim.islist or vim.tbl_islist ---@diagnostic disable-line: deprecated
+end
```
baod-rate, line 14: I'm also deprecating these functions in my code and using this as a reference. am I missing something here? or should this be:

```suggestion
---@param t? table
---@return boolean `true` if list-like table, else `false`.
function M.islist(t)
	return (vim.fn.has("nvim-0.10") == 1 and vim.islist or vim.tbl_islist)(t) ---@diagnostic disable-line: deprecated
end
```

or maybe just:

```suggestion
M.islist = vim.fn.has("nvim-0.10") == 1 and vim.islist or vim.tbl_islist
```

williamboman (commented):
Thanks for the PR! It seems like Neovim nightly temporarily broke due to the introduction of some calls that are not allowed in `:h vim.in_fast_event()` (in which Mason executes quite a lot of code). It seems like that issue was fixed in the next nightly release.

There are some issues with these changes currently - current usage of flatten expects support for keyed tables, the `islist()` function seems to return the function it's meant to call, and the introduction of functions in the `utils.lua` is a bit misplaced (this is a util module local to the `mason-core.installer.registry` module).

Thanks for brining it to my attention, I'll close this PR and open another one to replace these usages!

PR #1696, opened 2024-05-02
feat(registry): replace yq exec in file: source

baodrate:
for the file: registry source, replace the call to `yq` with a vendored pure-lua yaml parser.

This reduces our dependencies on external programs, especially since `yq` is not easily available on every system. e.g. the [default `yq` program in Arch](https://archlinux.org/packages/extra/any/yq/) is [kislyuk/yq](https://github.com/kislyuk/yq) rather than [mikefarah/yq](https://github.com/mikefarah/yq) as command seems to expect.

I confirmed by comparing the json output of the `github:mason-org/mason-registry` source against the json-encoded output of the file: source pointing at a local copy. Other than the order of some fields they are exactly the same

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/lua/mason-vendor/** linguist-vendored
--- a/lua/mason-registry/sources/file.lua
+++ b/lua/mason-registry/sources/file.lua
@@ -73,10 +73,7 @@ end
 function FileRegistrySource:install()
     return Result.try(function(try)
         a.scheduler()
-        if vim.fn.executable "yq" ~= 1 then
-            return Result.failure "yq is not installed."
-        end
-        local yq = vim.fn.exepath "yq"
+        local tinyyaml = require "mason-vendor.tinyyaml"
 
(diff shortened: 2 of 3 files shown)


williamboman (commented):
Thanks for the PR! I'd definitely prefer a no-dependency variant of this. I don't believe I ever tried tinyyaml when initially adding this capability, all other options I tried failed to properly parse simple YAML so I decided to offload it to `yq`.

I'm unable to find details on exactly what "subset" of YAML this implementation supports. I see you've modified the upstream code to support multi-line strings. Before moving to this implementation I'd like to better understand its limitations to avoid situations where the registry changes in such a way that it no longer is parseable. Do you happen to know what it doesn't support? I tried a couple of YAML documents from different sources and the only issue I discovered so far was multi-document files.

I'd also like to add that the `file:` protocol is so far only really added to aid in testing registries and is not really meant to be used "permanently".

williamboman:
Closing this as vendoring a yaml parser with unclear spec support is not of interest at the moment (although an interesting option).

PR #1703, opened 2024-05-11
fix: fix usage of deprecated Neovim APIs

williamboman:


erogers88:
@williamboman 

This change broke my configuration. It seems that I was using this "deprecated" API, the vim.tbl_keys() in mason_lspconfig.setup()'s ensure_installed parameter, and I got an error when this happened.

```
      local servers = {

        lua_ls = {
          Lua = {
            workspace = { checkThirdParty = false },
            telemetry = { enable = false },
          }
        },
        pyright = {},
        bashls = {}
      }

      mason_lspconfig.setup {
        ensure_installed = vim.tbl_keys(servers),
        ensure_installed = servers,
      }
```

Error Message
```
[mason-lspconfig.nvim] Server "pyright" is not a valid entry in ensure_installed. Make sure to only provide lspconfig se
rver names.
```


I was able to fix it by passing in a normal lua table:

```
      mason_lspconfig.setup {
        ensure_installed = servers,
      }
```

Should we let people know that their configuration might be broken now because of this? Maybe you can add it into the release notes? Also, how can I make myself aware of when things like this get deprecated?

Thanks for all your work, I love Mason by the way.

PR #1707, opened 2024-05-14
fix: remove the package from outdated_packages after uninstalling it

lnkkerst:
Remove the package from outdated_packages after uninstalling it. So it won't be reinstalled when updating all packages.

before:

https://github.com/williamboman/mason.nvim/assets/42333826/da84404b-abb8-4ba3-9d50-835f6c414e08

after:

https://github.com/williamboman/mason.nvim/assets/42333826/fd98a38c-174d-45cc-849d-a848db3ba836

### DIFF
--- a/lua/mason/ui/instance.lua
+++ b/lua/mason/ui/instance.lua
@@ -421,6 +421,9 @@ local function uninstall_package(event)
     ---@type Package
     local pkg = event.payload
     pkg:uninstall()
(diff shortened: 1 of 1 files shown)


williamboman:
Thanks for the PR! I saw this when you created the PR but I had already fixed it in the v2 release which I was still working on at that time. Sorry for not getting back to you sooner.

PR #1740, opened 2024-06-21
fix(powershell) convert "Stop" to 'Stop' and a join parameter to single quotes

thomastan:
Fixes http://github.com/williamboman/mason.nvim/issues/1739

sshot3-fix

A single-quote was needed to pass the 'Stop' through to `powershell`/`pwsh` correctly.

'SilentlyContinue' was single-quoted while "Stop" was double-quoted.  Converting "Stop" → 'Stop'.


williamboman (approved):
Thanks! I spent quite a lot of time trying to debug this but were unable to actually confirm whether the quotation marks are the root cause. I believe there's something else going on, but I guess it doesn't hurt to use single quotes everywhere.

PR #1759, opened 2024-07-20
feat: add backdrop

vuxuanhungg:
`Lazy` plugin manager creates a backdrop right behind its main floating window. This helps to draw our attention to the main window. It would be nice if `mason` could adopt this idea.

### Before:
image

### After:
Screenshot from 2024-07-20 20-33-45

As you can see, the backdrop brings depth to the main floating window, thus creates a better UI/UX.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,9 @@ local DEFAULT_SETTINGS = {
         -- The border to use for the UI window. Accepts same border values as |nvim_open_win()|.
         border = "none",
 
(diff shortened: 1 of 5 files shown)


williamboman (approved):
Thanks, looks cool! I took the opportunity to change the logic for how it closes the windows and buffers, also added some comments since the code is already pretty messy (and old).

Also fixed so that the backdrop is resized properly when the `VimResized` autocmd is fired.

PR #1781, opened 2024-09-04
feat(nuget): add download option to source and implement repository_url

616b2f:
Hi,

A little bit of context for the PR:

I was looking into a way to be able to install a "regular" nuget package to be able to install the official roslyn LSP from Microsoft which is not hosted on nuget.org (for details see https://github.com/dotnet/roslyn/issues/71474#issuecomment-2177303207).

The way currently nuget install works is we use "dotnet tool" to install nuget packages, but this works only with nuget packages which are "tools nuget packages" which means they have a specific structure and are for that also marked as tools in nuget repositories. When you try to install "regular" nuget packages via "dotnet tool" cli, it fails because the structure of the package is a different and it expect it to be formatted as a tool.

In this PR I introduced two things that allow us to install any nuget package if the "package" author decides to do that:

- possibility to describe we want only download the nuget package as is and extract it (because nupkg are just zip files)
- possibility to provide an alternative "repository_url" in the packages purl (which I implemented to work with dotnet tool and the new download way)

After this PR we should be able to do this PR (https://github.com/mason-org/mason-registry/pull/6330) for roslyn LSP in a more convenient way and more importantly to use the origin source for the LSP from the roslyn team.

This is my first PR here so please feel free to comment what I can improve or change.

PS: if possible we should squash the PR, because some commits where for a different solution initially which I abandoned when I came up with the current solution, but I thought it maybe helpful to keep it in the PR history.

### DIFF
--- a/lua/mason-core/installer/managers/nuget.lua
+++ b/lua/mason-core/installer/managers/nuget.lua
@@ -1,5 +1,6 @@
 local Result = require "mason-core.result"
 local installer = require "mason-core.installer"
+local log = require "mason-core.log"
 local platform = require "mason-core.platform"
 local fetch = require "mason-core.fetch"
 
@@ -11,8 +12,9 @@ local M = {}
 ---@param repository_url string
 ---@nodiscard
 function M.install(package, version, repository_url)
+    log.fmt_debug("nuget: install %s %s", package, version)
     local ctx = installer.context()
-
+    ctx.stdio_sink.stdout(("Installing nuget package %s@%s…\n"):format(package, version))
     local args = {
         "tool",
         "update",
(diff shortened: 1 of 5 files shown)


616b2f:
The PR is ready for review, if you have questions or need me to adjust anything I am happy to do so :)

616b2f:
I will look into the new command `dotnet nuget install` implemented into dotnet cli, we may simplify the logic here. See:

https://github.com/NuGet/Home/issues/12513

UPDATE: `dotnet nuget install` is not available in current sdk 10 LTS version so don't expect it to solve our problem in near future unfortunately.

So the PR stands with current solution.

616b2f:
I just realized that we need an issue first before I should create a PR, so I just created one. 

resolves #2045

616b2f:
Microsoft provides an official way to install the language server via `dotnet` cli as a nuget tool package via nuget.org, so this is not needed anymore as it is now.

Issue #1780, opened 2024-09-03
Listing installed packages is not working


Captainfl4me:
### I've searched open issues for similar requests

- [X] Yes

### I've recently downloaded the latest plugin version of mason.nvim

- [X] Yes

### Problem description

I was using my neovim config without issue on my previous Arch linux. Recently, I decided to reinstall in on my laptop and go through the installation process again, then pull my dotfiles config from my github so I dont' have to reconfigure everything. The problem is this time, whenever I try installing a LSP (either manually or via the ensure_installed list of mason-lsp) it works fine but then, when I restart neovim Mason cannot locate the previously installed package. I look for a solution online but never find it. Then, I deep dive into the code and activate the TRACE level for logs and find out that the function that scan the root dir was not working because the fs_readdir function didn't display a type inside the result object. After tracking down the issue I found this (https://github.com/luvit/luv/issues/660) which seems highly related. As it is a behavior that can happen maybe it would be interesting to take it into account when using the fs_readdir and make a case where the type field is nil.

### Expected behavior

Package install without error and are located after a restart

### Affected packages

All

### Installation log

```Text
[TRACE Wed 04 Sep 2024 12:16:10 AM CEST] .../.local/share/nvim/lazy/mason.nvim/lua/mason-core/fs.lua:20: fs: file_exists /home/nicoth/.local/state/nvim/mason-registry-update
[TRACE Wed 04 Sep 2024 12:16:10 AM CEST] .../.local/share/nvim/lazy/mason.nvim/lua/mason-core/fs.lua:11: fs: fstat /home/nicoth/.local/state/nvim/mason-registry-update
[TRACE Wed 04 Sep 2024 12:16:10 AM CEST] .../.local/share/nvim/lazy/mason.nvim/lua/mason-core/fs.lua:114: fs: read_file /home/nicoth/.local/state/nvim/mason-registry-update
[TRACE Wed 04 Sep 2024 12:16:10 AM CEST] .../.local/share/nvim/lazy/mason.nvim/lua/mason-core/fs.lua:20: fs: file_exists /home/nicoth/.local/share/nvim/mason/registries/github/mason-org/mason-registry/registry.json
[TRACE Wed 04 Sep 2024 12:16:10 AM CEST] .../.local/share/nvim/lazy/mason.nvim/lua/mason-core/fs.lua:11: fs: fstat /home/nicoth/.local/share/nvim/mason/registries/github/mason-org/mason-registry/registry.json
[...]

PR #1783, opened 2024-09-04
Fix/1780 more robust fs

Captainfl4me:
This pull request aims to fix this issue #1780. This issue is create by the libuv that sometimes can return nil type, therefore breaking the directory filter after a fs_readdir in the mason file system wrapper. To mitigate that I used the same fix as in [this issue](https://github.com/luvit/luv/issues/660) and use the fs_stat function if the type is not find by the fs_readdir function.


williamboman (approved):
Thanks!

PR #1876, opened 2025-02-13
Fixed deprecated calls of vim.validate

marks0mmers:
Hello, this is a simple PR to change the way `vim.validate` is called because of a deprecation in Neovim

This closes https://github.com/williamboman/mason.nvim/issues/1875


williamboman (approved):
Thanks! I added a check to make sure it only uses the new validate API when it's available.

PR #1894, opened 2025-03-15
Install ruby binaries with `--env-shebang`

nulty:
Using nvim on different projects with different ruby versions is made harder when I have to go to the Mason installed binaries and update the shebang so the formatter/lsp/whatever works across ruby versions.

Passing the `--env-shebang` options allows me to run the package using the currently enabled ruby versions.

### DIFF
--- a/lua/mason-core/installer/managers/gem.lua
+++ b/lua/mason-core/installer/managers/gem.lua
@@ -20,6 +20,7 @@ function M.install(pkg, version, opts)
         "install",
         "--no-user-install",
         "--no-format-executable",
+        "--env-shebang",
         "--install-dir=.",
         "--bindir=bin",
         "--no-document",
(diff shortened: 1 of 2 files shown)


phantomwhale:
Have moved onto this fork, and can confirm the strange issues I was having using Mason + chruby went away after using this fix 👌

phantomwhale:
I'll admit, my knowledge in this area is a quite hazy. But `chruby` works by shifting the PATH env to point at the correct ruby version (no shims) so I'm guessing using `env` binstubs means you'll always be picking up the appropriate version of ruby based on how the PATH has been set?

Perhaps the fundamental approach of `rbenv` requires different approaches here (?) Not sure. Might explain why I frequently have strange ruby issues around Mason tooling that I would have expected to be much more "visible" (I'm guessing more people still use "shim" approaches, such as rbenv, asdf or mise etc... than are using chruby?)

PR #1900, opened 2025-03-23
feat(window): support vim.o.winborder

iguanacucumber:
vim.o.winborder is an option that: Defines the default border style of floating windows. The default value is empty, which is equivalent to "none"

This PR adds support for this option and fixes bugs related to it


chrisgrieser:
Version 2.0, which has been released today, already addresses this issue, making this obsolete. Nonetheless, thanks!

Issue #1615, opened 2024-01-30
Installation of LSPs fails because directory already exists


nova-ace:
### I've searched open issues for similar requests

- [X] Yes

### I've recently downloaded the latest plugin version of mason.nvim

- [X] Yes

### Problem description

Installation of LSP packages fails during rename of .../staging/<package> to .../packages/<package>. This may be a unique problem to the filesystem that I am on in Rocky Linux 9 (BeeGFS). I was able to workaround the issue by commenting out line 235 in mason-core/instealler/context.lua:

-- fs.async.mkdir(install_path)

This occurs with any package I attempt to install not just lua-language-server

### Expected behavior

Possibly check for the above rename failure and revert to a copy operation. 

### Affected packages

All

### Mason output

```Text
None captured. Ran nvim MasonInstall in headless for troubleshooting.
```

### Installation log

```Text
[INFO  Tue 30 Jan 2024 04:15:17 PM MST] ...e/nvim/lazy/mason.nvim/lua/mason-core/installer/init.lua:184: Executing installer for Package(name=lua-language-server) {}
[DEBUG Tue 30 Jan 2024 04:15:17 PM MST] ...e/nvim/lazy/mason.nvim/lua/mason-core/installer/init.lua:58: Attempting to lock package Package(name=lua-language-server)
[DEBUG Tue 30 Jan 2024 04:15:17 PM MST] ...e/nvim/lazy/mason.nvim/lua/mason-core/installer/init.lua:70: Wrote lockfile Package(name=lua-language-server)
[DEBUG Tue 30 Jan 2024 04:15:17 PM MST] .../.local/share/nvim/lazy/mason.nvim/lua/mason-core/fs.lua:71: fs: mkdirp /mnt/beegfs/home/jake/.local/share/nvim/mason/staging/lua-language-server
[DEBUG Tue 30 Jan 2024 04:15:17 PM MST] ...zy/mason.nvim/lua/mason-core/installer/registry/init.lua:149: Compiling installer. lua-language-server {}
[DEBUG Tue 30 Jan 2024 04:15:17 PM MST] .../.local/share/nvim/lazy/mason.nvim/lua/mason-core/fs.lua:71: fs: mkdirp /mnt/beegfs/home/jake/.local/share/nvim/mason/staging/lua-language-server/libexec
[...]

PR #1933, opened 2025-05-08
Fix for issue #1615

Sam-Briney:
Fixes issue #1615 so that directory is renamed properly on beegfs

### DIFF
--- a/lua/mason-core/fs.lua
+++ b/lua/mason-core/fs.lua
@@ -79,6 +79,12 @@ local function make_module(uv)
     end
 
     ---@param path string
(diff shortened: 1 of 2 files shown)


williamboman (commented):
Hey and thanks for the PR! Do you get any specific error message, like `EEXIST`, `ENOENT`, `ENOTEMPTY`? I'm a bit careful with changing core components because they've undergone a lot of testing on different platforms, but obviously something seems to be wrong here.

Renaming to an existing, empty, directory should be supported by the `rename(2)` syscall, which libuv uses, so not sure if there's something else going on..

PR #1945, opened 2025-05-13
docs: make clear that `setup` is required

chrisgrieser:
There have been various issues in the past where ppl didn't call setup and have them reported that mason isn't working. This PR should help avoid that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,10 @@ your personal usage, some of these will also need to be installed. Refer to `:ch
 
 ```lua
 use {
-    "mason-org/mason.nvim"
+    "mason-org/mason.nvim",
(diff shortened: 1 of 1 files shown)


williamboman (commented):
I've actually wanted to remove the install section. The original idea behind keeping the snippets in the install section simple is because I've wanted to let people decide themselves how they set things up.

If people need to be able to copy paste a plugin spec, maybe it could be moved to the Setup section instead as a "recommended setup". It's been a recurring issue that people are setting up the plugin multiple times, for example via `opts = {}` and then random `require("mason").setup()` somewhere else. I'll open another PR as an alternative suggestion

Issue #1713, opened 2024-05-19
Support ensure_installed directly in mason


Limero:
### I've searched open issues for similar requests

- [X] Yes

### Is your feature request related to a problem? Please describe.

As a follow-up to https://github.com/williamboman/mason.nvim/issues/483 which adds the `ensure_installed` option to `mason-lspconfig`. It would be great if the same functionality could be added to mason itself. I prefer to set up my lsps manually without the use of `lspconfig` and would want mason to just automatically install them, nothing else.

### Describe the solution you'd like

`ensure_installed` being an option on `mason.nvim` without requring the use of `mason-lspconfig.nvim`.

To support the following:
```
require("mason").setup {
    ensure_installed = {
      "lua_ls",
      "rust_analyzer",
    },
}
```

ryuheechul:
for me, the motivation for this feature is to ensure installed for things other than lsp, like formatter, dap, linters, etc. since there is support for `ensure_installed` at https://github.com/williamboman/mason-lspconfig.nvim but I couldn't find anything similar for formatter, etc.

ionztorm:
any progress on this one at all?

PR #1949, opened 2025-05-13
feat(ensure_installed): add ensure_installed option

andreihh:
Fixes https://github.com/mason-org/mason.nvim/issues/1713.

Tested locally and it works:
```lua
{
  "andreihh/mason.nvim",
  branch = "ensure_installed",
  opts = { ensure_installed = { "lua-language-server@3.13.0", "stylua" } },
},
```

Implementation based on https://github.com/mason-org/mason-lspconfig.nvim/blob/main/lua/mason-lspconfig/features/ensure_installed.lua.

### DIFF
--- a/lua/mason/init.lua
+++ b/lua/mason/init.lua
@@ -27,17 +27,20 @@ function M.setup(config)
         Registry.sources:append(registry)
     end
 
-    local commands = require "mason.api.command"
+    local Command = require "mason.api.command"
     setup_autocmds()
 
-    local pkg_names_to_install = {}
-    for _, pkg_name in ipairs(settings.current.ensure_installed or {}) do
-        if not Registry.is_installed(pkg_name) then
-            table.insert(pkg_names_to_install, pkg_name)
+    local Package = require "mason-core.package"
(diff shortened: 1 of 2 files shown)


williamboman (commented):
Hello and thanks for the PR. This feature hasn't been added yet because:

1) As a principle mason.nvim tries to not provide any mechanisms for automatic installations of packages because there's a risk of misuse. All interactions (installing a package for example) needs to be done explicitly by the user. The Lua API exists to allow users to implement this themselves if they want to.
2) This implementation is problematic for two reasons:
    1) Performance. I haven't actually checked it but I'd not be surprised if the amount of modules loaded on setup is increased by 4x-5x. It'd also involve some minor overhead related to parsing and loading the registry. This is imo very unnecessary for a feature that realistically is only needed a handful of times.
    2) The semantics around version control of packages is incomplete. If a version is provided it only applies if the package is uninstalled - it doesn't for example pin the package to said version. It'd also interfere with other potential features like a lockfile.

As you can see it's a pretty simple feature to add. If you want to have something like this you can easily add it to your own Neovim config

PR #1951, opened 2025-05-15
docs: rework Installation & Setup sections, and other minor adjustments

williamboman:
Preview: https://github.com/mason-org/mason.nvim/tree/docs/setup.

@chrisgrieser What do you think?


README.md
```diff
@@ -39,16 +38,36 @@ with only a small set of [external requirements](#requirements) needed.
...
+
+`mason.nvim` is optimized to load as little as possible during setup. Lazy-loading the plugin, or somehow deferring the
+setup, is not recommended.
+
+Refer to the [Configuration](#configuration) section for information about which settings are available.
+
+## Recommended setup for `lazy.nvim`
+
+The following is the recommended setup when using `lazy.nvim`. It will set up the plugin for you, meaning **you don't have
+to call `require("mason").setup()` yourself**.
```
chrisgrieser, line 63: Is calling`setup` multiple times an issue for mason? If so, this could simply be fixed programmatically, by setting a `masonWasAlreadySetup` variable that disallows running `setup` twice. mason is one of the first plugins new users install, so keeping it as fool-proof as possible might make sense.

Alternatively, mason could also support changing the after when `setup` is called. Though that's not a hard standard, it is a pretty common practice among (newer) nvim plugins to allow the user to change a config by running `setup` again. Though this take a bit more effort, so disallowing multiple setup calls might be the simpler approach.

README.md
```diff
@@ -20,13 +20,12 @@
...
 -   [Requirements](#requirements)
--   [Installation](#installation)
--   [Setup](#setup)
-    -   [Extensions](#extensions)
 -   [Commands](#commands)
 -   [Registries](#registries)
+-   [Screenshots](#screenshots)
 -   [Configuration](#configuration)
 
 # Introduction
```
chrisgrieser, line 31: Just a nit since we are already reworking the docs: In Markdown, there is the convention to only use one h1 per file: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md001---heading-levels-should-only-increment-by-one-level-at-a-time

README.md
```diff
@@ -39,16 +38,36 @@ with only a small set of [external requirements](#requirements) needed.
 
 Packages are installed in Neovim's data directory ([`:h standard-path`][help-standard-path]) by default. Executables are
 linked to a single `bin/` directory, which `mason.nvim` will add to Neovim's PATH during setup, allowing seamless access
-from Neovim builtins (shell, terminal, etc.) as well as other 3rd party plugins.
+from Neovim builtins (LSP client, shell, terminal, etc.) as well as other 3rd party plugins.
 
 For a list of all available packages, see <https://mason-registry.dev/registry/list>.
```
chrisgrieser, line 43: could be a pretty link (`[]()`)

README.md
```diff
@@ -146,9 +122,16 @@ Before any packages can be used, the registry needs to be downloaded. This is do
 different Mason commands (e.g. `:MasonInstall`), but can also be done manually by using the `:MasonUpdate` command.
 
 If you're utilizing Mason's Lua APIs to access packages, it's recommended to use the
-[`:h mason-registry.refresh()`][help-mason-registry-refresh] and/or [`:h mason-registry.update()`][help-mason-registry-update]
+[`:h mason-registry.refresh()`][help-mason-registry-refresh] or [`:h mason-registry.update()`][help-mason-registry-update]
 functions to ensure you have the latest package information before retrieving packages.
 
+## Screenshots
```
chrisgrieser, line 128: should be same heading-level as the rest, since this is not a subsection of "Registries"

README.md
```diff
@@ -85,49 +104,6 @@ your personal usage, some of these will also need to be installed. Refer to `:ch
...
-setup, is not recommended.
-
-Refer to the [Configuration](#configuration) section for information about which settings are available.
-
-## Extensions
-
-Refer to the [Wiki](https://github.com/mason-org/mason.nvim/wiki/Extensions) for a list of 3rd party extensions.
-
--   [`mason-lspconfig.nvim`](https://github.com/mason-org/mason-lspconfig.nvim) - recommended for usage with `lspconfig`
-
```
chrisgrieser, line 130: This block is completely removed in the rewrite. Is that intentional? (Just checking in case you forgot to paste it after cutting.)

chrisgrieser (commented):
I noted two nits I noticed along the way, and thought they are worth a mention when rewriting the docs. But they absolutely not necessary minor stuff.

Also, do the `---@since 1.0.0` could maybe be removed? Not from the code (I assume they serve some automated docgen purpose), but since we are now at v2, such references to v1 feel a bit outdated.

README.md
williamboman: I kinda like the link being fully visible because it leads to an external domain

README.md
williamboman: Yeah this removal was intentional. `mason-lspconfig.nvim` isn't really as necessary anymore, and I kinda want to remove the idea of a "Mason ecosystem". I see a lot of people being confused about what the Mason plugins does and they end up assuming that either Mason is a requirement for LSP etc. or that `mason-lspconfig.nvim` is a requirement for usage with `nvim-lspconfig`, neither of which is true.

williamboman:
> Also, the `---@since 1.0.0` in the default config could maybe be removed? Not from the code (I assume they serve some automated docgen purpose), but in the README. Since we are now at v2, such references to v1 feel a bit outdated.

They primarily serve documentation purposes, there's no docgen involved atm. I kinda like having them visible in the docs. The documentation for the default configuration is done very lazily atm, just copy pasting the actual Lua table as-is.

README.md
chrisgrieser: Then how about using `has_setup` in `mason.nvim` as well to prevent multiple calls? Would be the simplest solution

chrisgrieser:
> > Also, the `---@since 1.0.0` in the default config could maybe be removed? Not from the code (I assume they serve some automated docgen purpose), but in the README. Since we are now at v2, such references to v1 feel a bit outdated.
> 
> They primarily serve documentation purposes, there's no docgen involved atm. I kinda like having them visible in the docs. The documentation for the default configuration is done very lazily atm, just copy pasting the actual Lua table as-is.

In principle I like such extra info, going through the releaes, release 1.0.0 is more than 2 years old. So the "since 1.0.0" info is only relevant to people who have not upgraded their mason in more than 2 years. 😅

williamboman:
> In principle I like such extra info, going through the releaes, release 1.0.0 is more than 2 years old. So the "since 1.0.0" info is only relevant to people who have not upgraded their mason in more than 2 years. 😅

Yeah it's probably not relevant anymore for such old versions. I think the problem is with how it's displayed as a simple Lua table with Lua comments. Also not sure where one would draw the line in terms of when to display it and when not to.

Anyways, I'll merge this in the current state

Issue #1961, opened 2025-05-23
Can't find absolute path to `npm` executable on PATH (Windows)


nataziel:
### I've searched open issues for similar requests

- [x] Yes

### I've recently downloaded the latest plugin version of mason.nvim

- [ ] Yes

### Problem description

Not 100% sure when this started occurring, but I just noticed this after updating `node` (installed via `scoop`). mason healthcheck reports being unable to find the absolute path to the npm executable in $PATH. The absolute path `C:/Users/me/scoop/apps/nodejs/current/npm` is NOT in $PATH, but the parent directory (`C:/Users/me/scoop/apps/nodejs/current`)  is.

``` bash
> which npm && where npm
/c/Users/me/scoop/apps/nodejs/current/npm
C:\Users\me\scoop\apps\nodejs\current\npm
C:\Users\me\scoop\apps\nodejs\current\npm.cmd
```

I think it's due to this change: https://github.com/mason-org/mason.nvim/pull/1885

### Expected behavior

Mason correctly finds the `npm` executable 

### Steps to reproduce

1. Install `npm` via `scoop install nodejs`
2. run `:checkhealth mason`

### Neovim version (>= 0.10.0)

NVIM v0.11.1
Build type: Release
LuaJIT 2.1.1741730670

### Operating system/version

Windows 11

### Healthcheck output

```shell
==============================================================================
mason:                                         require("mason.health").check()

mason.nvim ~
- ✅ OK mason.nvim version v2.0.0
- ✅ OK PATH: prepend
- ✅ OK Providers: 
    mason.providers.registry-api
    mason.providers.client
- ✅ OK neovim version >= 0.10.0

mason.nvim [Registries] ~
- ✅ OK Registry `github.com/mason-org/mason-registry version: 2025-05-22-hearty-tapir` is installed.

mason.nvim [Core utils] ~
- ✅ OK unzip: `UnZip 6.00 of 20 April 2009, by Info-ZIP.  Maintained by C. Spieler.  Send`
- ✅ OK wget: `GNU Wget 1.21.4 built on mingw32.
`
- ✅ OK curl: `curl 8.12.1 (Windows) libcurl/8.12.1 Schannel zlib/1.3 WinIDN
`
- ✅ OK gzip: `gzip 1.13`
- ✅ OK tar: `tar (GNU tar) 1.35`
- ✅ OK pwsh: `7.5.1 Microsoft Windows 10.0.22631 Win32NT
`
[...]

PR #1978, opened 2025-06-20
Fix path resolution issue on windows

unclenull:
Many windows tools are installed as scripts with extensions as `.bat` or `.cmd` on Windows.
Make this canonical and bulletproof.
#1961

### DIFF
--- a/lua/mason-core/spawn.lua
+++ b/lua/mason-core/spawn.lua
@@ -65,12 +65,13 @@ setmetatable(spawn, {
 
             -- Find the executable path via vim.fn.exepath on Windows because libuv fails to resolve certain executables
             -- in PATH.
-            if platform.is.win and (spawn_args.env and has_path(spawn_args.env)) == nil then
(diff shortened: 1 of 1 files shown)


lua/mason-core/spawn.lua
```diff
@@ -65,12 +65,13 @@ setmetatable(spawn, {
 
             -- Find the executable path via vim.fn.exepath on Windows because libuv fails to resolve certain executables
             -- in PATH.
-            if platform.is.win and (spawn_args.env and has_path(spawn_args.env)) == nil then
+            if platform.is.win and (#canonical_cmd <= 4 or canonical_cmd:lower():sub(-4) ~= ".exe") then
```
williamboman, line 68: This condition seems very arbitrary to me, why only for command names `<= 4` characters long? Also, mason never calls `spawn` with `.exe` explicitly so the second condition is unnecessary.

williamboman:
Superseded by #1991

PR #1984, opened 2025-07-10
fix(border): only set border to none if winborder doesn't exist

seblyng:
Problem: Custom border characters in winborder was just added. It needs to be a list of characters. `vim.o.winbar` gives the string value with comma separated items which is invalid.

Solution: Just never explicitly set the border if winborder option exists. From the docs of winborder:

> 	Defines the default border style of floating windows. The default value
	is empty, which is equivalent to "none".

### DIFF
--- a/lua/mason/ui/instance.lua
+++ b/lua/mason/ui/instance.lua
@@ -740,12 +740,8 @@ end
 
 local border = settings.current.ui.border
 
-if border == nil then
-    if vim.fn.has "nvim-0.11" == 1 then
-        border = vim.o.winborder
-    else
-        border = "none"
-    end
+if border == nil and vim.fn.exists "&winborder" == 0 then
+    border = "none"
 end
 
 window.init {


williamboman:
Hey! I'm not sure I understand what the issue is, I can't really produce any errors when trying out the new custom `'winborder'` support. The docs for `nvim_open_win` also mention that `border: The string form accepts the same values as the 'winborder' option.`

seblyng:
Yes but I believe that is not entirely true that it supports exactly the same string values. It doesn't seem to support string values with comma.

However, it is also unnecessary to set it explicitly since `nil` defaults to winborder.

Sorry I am on mobile, so the error is a bit bad, but I get this error:

image

seblyng:
You can try this minimal init:

```lua
vim.o.winborder = "a,a,a,a,a,a,a,a"
vim.pack.add({ "https://github.com/mason-org/mason.nvim" })
require("mason").setup()
```

and then run `:Mason`. A workaround I have used is this in the ui config option:
```
{
    ui = {
        border = vim.o.winborder:find(",") and vim.opt.winborder:get() or vim.o.winborder,
    }
}
```

williamboman:
Ok I see now, am able to reproduce. This kinda feels like a bug upstream, I raised an issue at https://github.com/neovim/neovim/issues/35065, will wait and see what it results in.

seblyng:
Okay, would you be open to merging this anyway? It's not necessary to set the border since nil defaults to winborder. So there is no change in behavior by merging this

williamboman:
Yep it'd make sense to just delegate it to default `'winborder'` behavior. I was a bit confused by the condition for `border = "none"` assignment, had to review the change a bit closer first. Thanks!

PR #2002, opened 2025-08-02
feat(compiler): make `supported_platforms` a universal source field

williamboman:
Previously this field had to be handled separately in each source type
(currently only supported by `cargo`, `gem`, `pypi`). This
backwards-compatible change makes `supported_platforms` a universal
top-level field on the `source:` object, meaning it'll be parsed for
each source type.

PR to modify the schema: https://github.com/mason-org/registry-schema/pull/17.

### DIFF
--- a/lua/mason-core/installer/compiler/compilers/cargo.lua
+++ b/lua/mason-core/installer/compiler/compilers/cargo.lua
@@ -1,45 +1,37 @@
 local Result = require "mason-core.result"
 local _ = require "mason-core.functional"
 local providers = require "mason-core.providers"
-local util = require "mason-core.installer.compiler.util"
 
 local M = {}
 
 ---@class CargoSource : RegistryPackageSource
----@field supported_platforms? string[]
 
 ---@param source CargoSource
 ---@param purl Purl
 function M.parse(source, purl)
-    return Result.try(function(try)
-        if source.supported_platforms then
-            try(util.ensure_valid_platform(source.supported_platforms))
-        end
+    local repository_url = _.path({ "qualifiers", "repository_url" }, purl)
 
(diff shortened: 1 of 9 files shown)


chrisgrieser:
While you are at it: [this PR](https://github.com/mason-org/mason-registry/pull/7535) revealed a tangentially related problem, regarding support for less popular platforms (freebsd), in case it's an easy fix that you can just tack on?

Conarius (approved):
LGTM

Conarius:
> While you are at it: [this PR](https://github.com/mason-org/mason-registry/pull/7535) revealed a tangentially related problem, regarding support for less popular platforms (freebsd), in case it's an easy fix that you can just tack on?

I think for that issue we would just have to setup CI tests on the mason-registry side mainly and adjust the enum of platforms and targets. Idk about the side of mason.nvim itself though.

williamboman:
> While you are at it: [this PR](https://github.com/mason-org/mason-registry/pull/7535) revealed a tangentially related problem, regarding support for less popular platforms (freebsd), in case it's an easy fix that you can just tack on?

I don't have access to a FreeBSD system so I can't really test and verify how to best do the platform detection. It's probably sufficient to rely on libuv's `uv_os_uname()`, similar to OpenBSD, but I'd need someone to verify.

> I think for that issue we would just have to setup CI tests on the mason-registry side mainly and adjust the enum of platforms and targets. Idk about the side of mason.nvim itself though.

Ultimately it'd have to be implemented in the mason.nvim client as well. The registry and the client are decoupled, we could "easily" add FreeBSD support to the registry, then it's just a matter for the client to recognize that platform.

PR #2052, opened 2026-01-07
feat: add support for removal of packages from a registry

williamboman:
This adds support for removal of packages from any given registry.
Currently mason.nvim doesn't support this at all and throws an error
when trying to interact with the registry in any way while having a
removed package installed locally.

This ensures that removed packages are available both in the `:Mason` UI
as well as the public Lua APIs. These "synthesized" packages only
supports uninstallation, and metadata such as licenses, categories,
homepage, etc is not available.

### DIFF
--- a/lua/mason-core/installer/InstallHandle.lua
+++ b/lua/mason-core/installer/InstallHandle.lua
@@ -43,7 +43,7 @@ function InstallHandleSpawnHandle:__tostring()
 end
 
 ---@class InstallHandle : EventEmitter
----@field package AbstractPackage
+---@field public package AbstractPackage
 ---@field state InstallHandleState
 ---@field stdio_sink BufferedSink
 ---@field is_terminated boolean
--- a/lua/mason-core/installer/context/init.lua
+++ b/lua/mason-core/installer/context/init.lua
@@ -17,7 +17,7 @@ local receipt = require "mason-core.receipt"
 ---@field location InstallLocation
 ---@field spawn InstallContextSpawn
 ---@field handle InstallHandle
----@field package AbstractPackage
+---@field public package AbstractPackage
 ---@field cwd InstallContextCwd
 ---@field opts PackageInstallOpts
 ---@field stdio_sink StdioSink
--- a/lua/mason-core/package/AbstractPackage.lua
+++ b/lua/mason-core/package/AbstractPackage.lua
@@ -6,7 +6,6 @@ local Result = require "mason-core.result"
 local _ = require "mason-core.functional"
 local fs = require "mason-core.fs"
 local log = require "mason-core.log"
-local path = require "mason-core.path"
 local settings = require "mason.settings"
 local Semaphore = require("mason-core.async.control").Semaphore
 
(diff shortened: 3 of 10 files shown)


williamboman:
Demo https://asciinema.org/a/lIlo2OrZmYJ7GWLrDskhxqbqx

Conarius (approved):
LGTM

chrisgrieser:
From the perspective of the mason registry:
1. What kind of policy should we have on removing packages? Usually, "deprecated" means that something will be removed at some point in the future. I am pretty sure that by now, we have a bunch of packages that have been deprecated years ago. I think some rule of thumb like "a package should be deprecated for at least 6 months before removal" might make sense (time span of months up for discussion of course)
2. Just to make sure I understand this correctly, removing a package is now pretty much just deleting the `package.yaml` of the respective file, right?

williamboman:
> * What kind of policy should we have on removing packages? Usually, "deprecated" means that something will be removed at some point in the future. I am pretty sure that by now, we have a bunch of packages that have been deprecated years ago. I think some rule of thumb like "a package should be deprecated for at least 6 months before removal" might make sense (time span of months up for discussion of course)

I think 6 months would be a good timespan to adhere to. I think it also depends on the package in question, there might be packages that technically are deprecated but are still broadly used where it would be disruptive to remove it - but as a rule of thumb anywhere from 3-6 months sounds OK to me (I'm leaning towards 3 months).

In the deprecation entry, maybe we should use dates instead of versions in the `since` field to easily keep track of this. This date would be when the package was deprecated in the registry, not necessarily when the package actually was deprecated.

```yaml
deprecation:
  since: 2026-01-06
```

> * Just to make sure I understand this correctly, removing a package is now pretty much just deleting the `package.yaml` of the respective file, right?

Yep. To be technically specific, it requires the removal of the entire `package/<package-name>/` directory.

*Important: We should wait before we start removing packages from the registry so that this features reaches a vast majority of users first, otherwise people will quickly run into issues.*

chrisgrieser:
> I think 6 months would be a good timespan to adhere to. I think it also depends on the package in question, there might be packages that technically are deprecated but are still broadly used where it would be disruptive to remove it - but as a rule of thumb anywhere from 3-6 months sounds OK to me (I'm leaning towards 3 months).
>
> In the deprecation entry, maybe we should use dates instead of versions in the since field to easily keep track of this. This date would be when the package was deprecated in the registry, not necessarily when the package actually was deprecated.

Sounds both good. I think a PR to the mason-registry might be good, adding all this information to the contribution info and the specification of the `deprecated` field. Not only as a reference for us and future maintainers, but also as info/guideline for contributors.

> Yep. To be technically specific, it requires the removal of the entire package/<package-name>/ directory.

To be technically *more* specific, git tracks only files not folders, and since the `package.yaml` is the only file in the package directory, deleting that file effectively means deleting the folder as well. ;)

> *Important: We should wait before we start removing packages from the registry so that this features reaches a vast majority of users first, otherwise people will quickly run into issues.*

Good point. I've set a reminder for in 3 months to start removing deprecated packages from the registry.

williamboman:
FYI @deirn we'll likely start removing packages from the core registry in the coming months, not sure how your client handles that atm.

deirn:
It copies the installed package recipe to a local file, so it should be fine.

Edit: actually, checked it and it in fact does NOT support it, lol. Fixing it.

deirn:
Fixed in https://github.com/deirn/mason.el/commit/47b6e24820fb65bc934fca3ebc17d215e9db9893

PR #2081, opened 2026-05-03
chore(pypi): add python 3.13 and 3.14 to list of fallbacks

aauren:
Adds Python 3.13 and Python 3.14 to the list of Python fallback versions so that they are discoverable if the primary `python` or `python3` binary fails to install the package.

In my case I was trying to install `nginx-language-server` on an Arch Linux machine that only has 3.14 available. I installed 3.13 via the AUR, but it still couldn't install the package. After making this change, the install succeeded fine because it was able to find `python3.13` in the PATH.

### DIFF
--- a/lua/mason-core/installer/managers/pypi.lua
+++ b/lua/mason-core/installer/managers/pypi.lua
@@ -66,15 +66,16 @@ local function get_versioned_candidates(supported_python_versions)
         end
         return Optional.of(executable)
     end, {
-        { semver.new "3.14.0", "python3.14" },
-        { semver.new "3.13.0", "python3.13" },
-        { semver.new "3.12.0", "python3.12" },
-        { semver.new "3.11.0", "python3.11" },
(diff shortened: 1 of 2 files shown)


williamboman (approved):

aauren:
Hmm... I tried to parse all of the GH action output, but I can't find anything that says why the tests are failing. It looks like all of the tests pass, but then make returns 1 at the end of the run.

My change is so small, I can't see how adding these additional Python interpreters would cause these failures. But definitely open to making changes if necessary.

williamboman:
Had to debug the CI failures more than I'd like to admit 😅. Tests were timing out in a weird way due to mocks not being set up.

I also made a change where it'll try to find the lowest available python version instead of the highest available version. I remember this being an issue in the past where for example a package would specify support for `>=3.10` but in reality wouldn't be compatible with `python 3.14` and crash when executed.

PR #2088, opened 2026-05-16
feat: add support for socket.dev firewall client

williamboman:
Screenshot 2026-05-16 at 17 57 38
Screenshot 2026-05-16 at 17 57 43

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@
 -   [Commands](#commands)
 -   [Registries](#registries)
 -   [Screenshots](#screenshots)
+-   [Firewall (socket.dev)](#firewall-socketdev)
 -   [Configuration](#configuration)
 
 ## Introduction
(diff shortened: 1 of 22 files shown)


williamboman:
@mason-org/triage If you want to try this out I'd appreciate it! It should be fully seamless, but needs to manually be enabled:

```lua
require("mason").setup {
  firewall = {
    enabled = true
  }
}
```

Conarius (approved):
LGTM

williamboman:
A note on the "system packages" available in https://github.com/mason-org/mason-system-registry. The idea is for every package to include a version scope in the package name itself. In the case of `sfw` the package name is `sfw@latest`. Should more packages be available in the future, for example Python and Node, these would be separate packages named like:

- `node@18`
- `node@20`
- `node@latest`
- `python@3.12`
- `python@3.11`

PR #2095, opened 2026-05-28
Fixes #2094

### LINKED ISSUE #2094: Suggested Lazy.vim config errors out
The suggested configuration files for Lazy.vim produce errors like:

> Error detected while processing /home/user/.config/nvim/init.lua:
> Failed to load `plugins.mason`:
> /home/user/.config/nvim/lua/plugins/mason.lua:1: unexpected symbol near '{'

ldesousa:
Simple change to make Lazy.vim happy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ The following is the recommended setup when using `lazy.nvim`. It will set up th
 to call `require("mason").setup()` yourself**.
 
 ```lua
-{
+return {
     "mason-org/mason.nvim",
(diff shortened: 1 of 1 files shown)


williamboman:
Hey, this seems more like a very basic user error. The example config spec in the readme can be placed in different locations, most of which does not require the `return` keyword. I looked at a sample of other plugin readmes and it doesn't seem to be standard to include `return`.

